PR #1, opened 2021-08-23
Add SamuraiHudWindow

abazilla:
Had a go at adding Samurai to the list of HUDs! Please give me some feedback - it's both my first time in C# and contributing to open source :) 

gif: https://imgur.com/XN6bqJr

### DIFF
--- a/Interface/SamuraiHudWindow.cs
+++ b/Interface/SamuraiHudWindow.cs
@@ -1,0 +1,133 @@
+﻿using System;
+using System.Numerics;
+using Dalamud.Game.ClientState.Structs.JobGauge;
+using Dalamud.Plugin;
+using ImGuiNET;
+
+namespace DelvUIPlugin.Interface {
+    public class SamuraiHudWindow : HudWindow {
+        public override uint JobId => 38;
+        
+        private new static int BarHeight => 26;
+        private new static int BarWidth => 357;
+        private new static int XOffset => 178;
+        private new static int YOffset => 496;
+        
+        public SamuraiHudWindow(DalamudPluginInterface pluginInterface, PluginConfiguration pluginConfiguration) : base(pluginInterface, pluginConfiguration) { }
+
+        protected override void Draw(bool _) {
+            DrawHealthBar();
+            DrawPrimaryResourceBar();
+            DrawSenResourceBar();
+            DrawMeditationResourceBar();
+            DrawTargetBar();
+        }
+
+        protected override void DrawPrimaryResourceBar()
+        {
+            var gauge = PluginInterface.ClientState.JobGauges.Get<SAMGauge>();
+
+            const int yPadding = 3;
+            var barWidth = BarWidth;
+            var xPos = CenterX - XOffset;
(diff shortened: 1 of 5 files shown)


Interface/SamuraiHudWindow.cs
```diff
@@ -0,0 +1,155 @@
...
+            DrawSenResourceBar();
+            DrawMeditationResourceBar();
+            DrawTargetBar();
+        }
+
+        protected override void DrawPrimaryResourceBar()
+        {
+            var gauge = PluginInterface.ClientState.JobGauges.Get<SAMGauge>();
+
+            // const int xPadding = 5;
```
jdsmith2816, line 30: Let's try to keep commented out code out of the repo;  everything is in source control so we can always bring it back if we need it.

Interface/SamuraiHudWindow.cs
```diff
@@ -0,0 +1,155 @@
...
+            
+            var barWidth = (BarWidth - xPadding * (numChunks - 1)) / numChunks;
+            var barSize = new Vector2(barWidth, BarHeight);
+            var xPos = CenterX - XOffset;
+            var yPos = CenterY + YOffset ;
+            var cursorPos = new Vector2(xPos - xPadding - barWidth, yPos);
+            
+            var drawList = ImGui.GetWindowDrawList();
+            
+            for (var i = 1; i < 4; i++) {
```
jdsmith2816, line 79: Why are you looping over this? I don't think we're gaining anything in clarity or brevity here. 

Personally I think we remove the loop and the `if i == #` checks and just have the duplicated cursorpos and border rect lines.

Interface/SamuraiHudWindow.cs
```diff
@@ -0,0 +1,155 @@
...
+                cursorPos = new Vector2(cursorPos.X + xPadding + barWidth, cursorPos.Y);
+                
+                if (i == 1)
+                {
+                    if (gauge.HasSetsu())
+                    {
+                        drawList.AddRectFilled(
+                        cursorPos, cursorPos + barSize,
+                        ImGui.ColorConvertFloat4ToU32(new Vector4(89 / 255f, 234 / 255f, 247 / 255f, 255f))
```
jdsmith2816, line 88: IMO we should just directly put the color here.   If you google `rgb to hex` you'll get a little widget with an RGB box on the bottom left that you can type `89 234 247` into and it'll show you the HEX value of `#59eaf7` 

Now that you know the hexcode you can replace this line of code with 0xAABBGGRR  AA is FF for 100% opacity so your color becomes.. 0xFFF7EA59

Repeat this process for the other three colors.

jdsmith2816 (changes requested):
This mostly looks fine to me other than the few comments I left.  Great first PR!

One final note... Let's try to keep the commitlog relatively clean.  When you update the PR for needed changes go ahead and squash down to one commit with a message like

`feat: add samurai support with custom job gauges`

abazilla:
All done! :) thanks for the review.

jdsmith2816:
Thanks abazilla!

PR #11, opened 2021-08-25
Add Dragoon HUD Window

asgoel:
The colors I'm iffy about, but this is what it looks like right now:

https://imgur.com/a/T1VWP4r

Also not sure about the .csproj folder, never worked in C# before.

### DIFF
--- a/DelvUI/Interface/ConfigurationWindow.cs
+++ b/DelvUI/Interface/ConfigurationWindow.cs
@@ -2777,6 +2777,85 @@ namespace DelvUI.Interface
 
                     ImGui.EndTabItem();
                 }
+                if (ImGui.BeginTabItem("Dragoon"))
+                {
+                    var eyeOfTheDragonHeight = _pluginConfiguration.DRGEyeOfTheDragonHeight;
+                    if (ImGui.DragInt("Eye of the Dragon Bar Height", ref eyeOfTheDragonHeight, .1f, 1, 1000))
+                    {
+                        _pluginConfiguration.DRGEyeOfTheDragonHeight = eyeOfTheDragonHeight;
+                        _pluginConfiguration.Save();
+                    }
+
+                    var eyeOfTheDragonBarWidth = _pluginConfiguration.DRGEyeOfTheDragonBarWidth;
+                    if (ImGui.DragInt("Eye of the Dragon Bar Width", ref eyeOfTheDragonBarWidth, .1f, 1, 1000))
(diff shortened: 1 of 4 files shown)


Interface/DragoonHudWindow.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        protected override void DrawPrimaryResourceBar()
+        {
+            var gauge = PluginInterface.ClientState.JobGauges.Get<DRGGauge>();
+
+            var barWidth = (BarWidth - xPadding * 2) / 2.0f;
+            var barSize = new Vector2(barWidth, BarHeight);
+            var xPos = CenterX - XOffset;
+            var yPos = CenterY + YOffset - 33;
```
VeraThorne, line 37: This 33 is a repeated value I notice, is there a constant that can be extracted here?

Interface/DragoonHudWindow.cs
```diff
@@ -0,0 +1,102 @@
...
+                drawList.AddRectFilled(cursorPos, cursorPos + barSize, 0xFFC2B6FF);
+            }
+            else
+            {
+                drawList.AddRectFilled(cursorPos, cursorPos + barSize, 0x88000000);
+            }
+
+            drawList.AddRect(cursorPos, cursorPos + barSize, 0xFF000000);
+
+            // Chunk 2
```
VeraThorne, line 54: These two chunks are identical beyond the eye count check number. Maybe it would be better to merge these into a for loop based on the count of chunks you want?

Interface/DragoonHudWindow.cs
```diff
@@ -0,0 +1,102 @@
...
+            var xPos = CenterX - XOffset;
+            var yPos = CenterY + YOffset - 33;
+            var cursorPos = new Vector2(xPos, yPos);
+            var eyeCount = gauge.EyeCount;
+
+            // Chunk 1
+            var drawList = ImGui.GetWindowDrawList();
+            if (eyeCount >= 1)
+            {
+                drawList.AddRectFilled(cursorPos, cursorPos + barSize, 0xFFC2B6FF);
```
VeraThorne, line 45: Is there somewhere that might be better suited for storing these colors? Since it seems like they're meant to be used multiple times might make future maintenance easier if these were constants (and later loaded in via configuration if the functionality is so desired)

asgoel:
@jdsmith2816 pinging on this one

PR #12, opened 2021-08-25
fixed inconsistant dancer bar sizing

TheResinger:
Updated the code for the padding and rendering of the two bars to be more inline with the bars bard and summoner uses.

Preview: https://i.imgur.com/xeXCP16.png


jdsmith2816 (approved):
LGTM 👍 :shipit:

PR #18, opened 2021-08-25
Adding support for BLM

Tischel:
image

Bar = Mana Bar (changes color depending on current element)
Bar Glow = Enochian active
Timer = Element timer
White Bars = Umbral Ice stacks
Pink Bars = Polyglot stacks - They fill up as you maintain Enochian (you get 1 charge every 30sec of Enochian being up)
White Dots = Triplecast stacks

### DIFF
--- a/Interface/BlackMageHudWindow.cs
+++ b/Interface/BlackMageHudWindow.cs
@@ -1,0 +1,190 @@
+﻿using System;
+using System.Numerics;
+using Dalamud.Game.ClientState.Structs.JobGauge;
+using Dalamud.Plugin;
+using ImGuiNET;
+using System.Linq;
+using System.Collections;
+
+namespace DelvUIPlugin.Interface
+{
+    public class BlackMageHudWindow : HudWindow
+    {
+        public override uint JobId => Jobs.BLM;
+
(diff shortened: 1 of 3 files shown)


Interface/BlackMageHudWindow.cs
```diff
@@ -0,0 +1,190 @@
...
+            var drawList = ImGui.GetWindowDrawList();
+
+            var positions = new ArrayList();
+            positions.Add(new Vector2(CenterX, CenterY + YOffset - 100));
+            positions.Add(new Vector2(CenterX - JobBarWidth / 2f + 30, CenterY + YOffset - 85));
+            positions.Add(new Vector2(CenterX + JobBarWidth / 2f - 30, CenterY + YOffset - 85));
+
+            for (int i = 0; i < 3; i++)
+            {
+                if (tripleStackBuff.StackCount <= i)
```
VeraThorne, line 181: Can just loop using the StackCount as a range.

```suggestion
               for (int i= 0; i< TripleStackBuff.StackCount; i++)
               {
                   drawList.AddCircle((Vector2)positions[i], 10, 0xFFFFFFFF, 6, 4);
               }
```

PR #41, opened 2021-08-28
add configuration for SMN

macb:
Defaults line up with the existing placement. Screenshot is with a 220 Y offset (also applied to general frames)

TODO:
- [x] Add color support

image


macb:
image
image

PR #46, opened 2021-08-28
Added option to toggle the flashing interrupt castbar on/off when you interrupt a cast

qwakk:


Kanubiz (changes requested):
Remove all changes to .csprog and .sln please

PR #52, opened 2021-08-28
Fix build when spaces in project path

Bedo9041:
With the previous project file, the build would fail if there were spaces in the project path. This fixes it, but does remove the catch-all that was used by ILRepack (though it shouldn't matter unless more dlls are going to be used for ILRepack in the future)


jdsmith2816:
Thanks Bedo1

PR #55, opened 2021-08-28
Generic bars

Bedo9041:
An initial implementation of generic bars, currently has the generic bar system and three jobs are using it. There are many more features that could be added to these bars, however most jobs as they are currently implemented should be able to swap over. 

As an example, this is how the MCH battery gauge is currently implemented:

https://github.com/jdsmith2816/delvui/blob/e1619d3ca1235d6ec773d217793d2cbd42ea1734/DelvUI/Interface/MachinistHudWindow.cs#L132-L223

This is how it is implemented in this PR:

https://github.com/Bedo9041/delvui/blob/2ff404f6137afea2e9fe6b88081e23f544e8855c/DelvUI/Interface/MachinistHudWindow.cs#L107-L129

Obviously as this would be a fairly large change to how bars are drawn, and it's not as robust as I would like (for example, right now you would likely have to fall back to manual rendering if you want more than one piece of text per bar 'chunk') I'm opening the PR as a draft mainly for feedback if this is how we want to do it.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/FFXIVClientStructs"]
+	path = lib/FFXIVClientStructs
+	url = https://github.com/aers/FFXIVClientStructs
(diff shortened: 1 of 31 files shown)


Kanubiz:
awesome!

poly1k:
This is rad!

PR #60, opened 2021-08-29
Mp ticker

Tischel:
Looks like this:
https://youtu.be/q2FZcKLTR7A

Also contains the refactor on the Draw function that we talked about in discord.


Kanubiz (changes requested):
Add border option

Kanubiz (changes requested):
Also, disable on full mp option (disabled by default)

PR #76, opened 2021-08-30
fix: aesthetician (barber) crash

DomGrieco:
There was a null pointer on the castbar draw while loading into the character customization screen in game. To circumvent this I decided to make a bool check in the draw function inside Plugin.cs. The bool checks the ClientState for specific condition flags. If any return true, the hud is not drawn.


DomGrieco:
Looks like I made a learning mistake with GitKraken and this is a duplicate. Closing.

PR #90, opened 2021-08-30
Fixing mana text offsets

Tischel:


Kanubiz (changes requested):
change the into to rdm-mana-text

PR #97, opened 2021-08-31
fix: Nvidia GPU crash (hopefully)

DomGrieco:
Also ran resharper on ConfiguraitonWindow.cs


Kanubiz:
Nvidia GPU crash (hopefully)

PR #160, opened 2021-08-31
feature (NIN): added a joint bar for trick attack and suiton

nilaykumar:
Added a joint bar for trick attack (red) and suiton (secondary, white).
image


poly1k:
This will conflict with #159.

Can you remove initialHeight and set good default offsets (bottom bar 1px above castbar etc.)?

PR #166, opened 2021-08-31
Fix crash on object interaction with purple quest ground marker

arejensen:
Interaction with quest item on purple ground marker objective would lead to null being passed to DrawOutlinedText and a memory access violation.

Now passes empty string if argument is null.

### DIFF
--- a/DelvUI/Interface/HudWindow.cs
+++ b/DelvUI/Interface/HudWindow.cs
@@ -443,7 +443,7 @@ namespace DelvUI.Interface {
 
             if (PluginConfiguration.ShowActionName) {
                 DrawOutlinedText(
-                    castText ??= "",
+                    castText,
                     new Vector2(
                         cursorPos.X + (PluginConfiguration.ShowActionIcon && _lastPlayerUsedCast.HasIcon ? CastBarHeight : 0) + 5,
                         cursorPos.Y + CastBarHeight / 2f - castTextSize.Y / 2f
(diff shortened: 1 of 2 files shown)


DelvUI/Interface/HudWindow.cs
```diff
@@ -443,7 +443,7 @@ protected void OnConfigChanged(object sender, EventArgs args)
 
             if (PluginConfiguration.ShowActionName) {
                 DrawOutlinedText(
-                    castText,
+                    castText ??= "",
```
jdsmith2816, line 446: Should just be `castText ?? ""` no assignment required, right?

jdsmith2816 (changes requested):
Should just be castText ?? "" no assignment required, right?

In either case I think that the better solution would be to fix this in `DrawOutlinedText` instead

Kanubiz:
I would decline this for now since we need the crashes to properly map the IDS

arejensen:
> Should just be `castText ?? ""` no assignment required, right?

You are absolutely right. 
 
> In either case I think that the better solution would be to fix this in `DrawOutlinedText` instead

I've added a temporary fix in `DrawOutlinedText` until we can figure out the issue with id mapping.

jdsmith2816:
Your fix sounds plausible but the `DrawOutlinedText` method should still not crash on null text.  We need that fixed in either case.

Zeffuro:
> Your fix sounds plausible but the `DrawOutlinedText` method should still not crash on null text. We need that fixed in either case.

Yes sanitycheck still needs to be there, agreed.

arejensen:
I've submitted #182 which incorporates a check on actor type as outlined by Zeffuro.

PR #175, opened 2021-09-01
Dancer Proc Bar

chipget:
This pull request adds bars for the single target DNC procs (Flourishing Cascade and Flourishing Fountain), which default to the color of their corresponding step transformation (i.e., the color of Flourishing Cascade defaults to that of Jete, Flourishing Fountain to Pirouette). The bars default to the size of the feather bars, use the same padding as the feather bars, and default to a location just above those feather bars.

### DIFF
--- a/DelvUI/Interface/ConfigurationWindow.cs
+++ b/DelvUI/Interface/ConfigurationWindow.cs
@@ -3822,6 +3822,13 @@ namespace DelvUI.Interface
                         _pluginConfiguration.Save();
                     }
 
+                    var procTimersEnabled = _pluginConfiguration.DNCProcTimersEnabled;
+                    if (ImGui.Checkbox("Proc Timers Enabled", ref procTimersEnabled))
+                    {
+                        _pluginConfiguration.DNCProcTimersEnabled = procTimersEnabled;
+                        _pluginConfiguration.Save();
+                    }
+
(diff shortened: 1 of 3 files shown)


chipget:
dnc_image

Bedo9041 (changes requested):
This appears to be missing Rising Windmill and Bloodshower procs. Ideally they would also be included in the proc bar.
Also this is potentially personal preference, but I'm not a fan of having multiple timer bars on the horizontal bar without them being symmetrical - but I don't think there's a good way to do this for four different procs without significantly cluttering the UI using multiple bars.

Bedo9041:
I wouldn't worry about the difference between AoE and ST procs, AoE procs are still used in ST fights over the standard 1-2 combo. And you still get the procs because of flourish.
Maybe have timers be an option, turned off by default - that way if people want the timers (which are useful for deciding between saber dance and using a proc if your getting a ton of esprit in a burst window) they can turn them on.

chipget:
Okay, this branch now includes your suggestions. The following images show the updated bars. The first corresponds to the fixed bar option, and the final one corresponds to the timer bar option.

dnc_cascade_fixed

dnc_fountain_fixed

dnc_procs_timers

dnc_shower_fixed

dnc_windmill_fixed

Bedo9041 (approved):

PR #191, opened 2021-09-01
Add `/pdelvui toggle` command to toggle HUD visibility

bayov:
Adds an ingame command to toggle HUD visibility:
image


bayov:
Note, it might be more convenient to add `/pdelvui show` and `/pdelvui hide` instead of toggling.

PR #199, opened 2021-09-02
Display MP Value and Treshold on Healer jobs

Incalex:
Currently, all healer jobs (WHM/SCH/AST) use the default PrimaryResourceBar to display their MP.

This modification adds the configuration to display the current MP Value and a threshold bar (like the already existing one for the RDM MP Bar).

I made the configuration modifications in the General tab as it is where the Primary Resource bar configuration is already present but currently, it only affects healer jobs and not currently managed jobs (BLU, DoH, DoL).

Example on AST:
image

Added configuration:
image


Kanubiz:
Good addition

Kanubiz (approved):

PR #225, opened 2021-09-03
Status effects refactor

Tischel:
https://youtu.be/li77IA1_exc

**Big refactor on Buffs / Debuffs.**

* Each list is an instance of StatusEffectsList with its own configuration
* Refactored how the configuration is saved. Instead of having a million fields on the PluginConfiguration class, I think we should start using serializable objects to use as config. In this PR you'll see a good example of that.
* Replaced row and columns settings to constrained area in preparation for draggable elements.
* Added a lot of settings for customization of the layout and icons
* Revamped config window to accommodate for the changes.


Zeffuro (commented):
Looks solid, very solid implementation

PR #242, opened 2021-09-03
Raid/Job Auras WIP

asgoel:


DelvUI/Interface/StatusEffects/StatusEffectsList.cs
```diff
@@ -126,15 +135,34 @@ private List<StatusEffectData> StatusEffectsData()
...
+            var toReturn = new List<StatusEffectData>();
+            foreach (uint buffId in filterBuffs)
+            {
+                var idx = list.FindIndex(s => (uint)s.statusEffect.EffectId == buffId);
+                if (idx >= 0)
+                {
+                    toReturn.Add(list[idx]);
+                }
+            }
+            return toReturn;
```
Tischel, line 152: Instead of iterating this after going trough the actor buffs, maybe it's better to do it directly in the for..loop right before.
`if (filterBuffs.Contains(status,EffectId) continue;`

Tischel (approved):

DelvUI/Interface/StatusEffects/StatusEffectsList.cs
asgoel: If you put it in the above loop, you can't guarantee that the resulting buff array follows the same priority as what's in filter buffs. 

Since this loop isn't nested it shouldn't affect the actual overall complexity of the function.

Tischel: You're totally right, didn't realize the filter buffs were also used to sort the list. Makes sene!

asgoel:
Let me fix up the merge conflicts, I guess

asgoel:
@Tischel this one is up to date with main now

Tischel:
@asgoel @Kanubiz PR looks good to me, need to point it to develop before merging though.

asgoel:
> @asgoel @Kanubiz PR looks good to me, need to point it to develop before merging though.

done

Tischel:
> > @asgoel @Kanubiz PR looks good to me, need to point it to develop before merging though.
> 
> done

I dont have permissions to merge so we'll have to wait for @Kanubiz

Kanubiz (approved):

PR #243, opened 2021-09-03
feat(ast): more options, new menu and Vector2

Vayhyr:
feat: added option to show cumulative cooldowns of Redraw (cooldown until next charge is now default)
feat: show Dots/Draw/Redraw/Star/Lightspeed with a decimal precision
feat: added option to hide text per bar
feat: new menu for AST
chore: swapped bar x/y and weight/width int to Vector 2

preview menu:
ast-menu-1

ast-menu-2

preview decimal bars:
decimal-bars


DelvUI/Helpers/Vector2Helper.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            return new Vector2(x, v.Y);
+        }
+
+        public static Vector2 ChangeY(Vector2 v, float y)
+        {
+            return new Vector2(v.X, y);
+        }
+    }
+}
```
Tischel, line 23: I don't see this being used.
If so please remove.

jdsmith2816: Unrelated but just in case it ever helps... You can also do vec2 + 1 and end up with the result of new Vector2(vec2.x + 1, vec2.y + 1)

Vector2.One and Vector2.Zero exist as do Vector2.UnitX and Vector2.UnitY

Vector2.UnitX * scale  would result in new Vector2(scale, 0)

Vayhyr: I removed it since it doesn't serve its purpose with the new config system in place.

DelvUI/Interface/AstrologianHudWindow.cs
```diff
@@ -1,121 +1,81 @@
...
         }
 
-        private void DrawDivinationBar() {
-            var chunkColors = new List<Dictionary<string, uint>>();
-
-            if (PluginInterface.ClientState.LocalPlayer?.ClassJob.Id != 33) {
+        private void DrawDivinationBar()
+        {
+            List<Dictionary<string, uint>> chunkColors = new List<Dictionary<string, uint>>();
+            if ((PluginInterface.ClientState.LocalPlayer?.ClassJob.Id) != 33)
```
Tischel, line 72: Curious on why you need to check the job id here.
Supposedly this is being taken care of before this class gets told to draw.
Also if this is really necessary, please replace the hardcoded number by using the Jobs enum.

DelvUI/Interface/AstrologianHudWindow.cs
```diff
@@ -313,52 +292,70 @@ internal class AstrologianHudWindow : HudWindow {
...
+                    ? ((!_config.ASTEnableRedrawCooldownCumulate
+                        ? ((redrawCastInfo % 30) == 0
+                            ? 30
+                            : (_config.ASTShowRedrawCooldownTextBar
+                                ? (redrawCastInfo % 30).ToString(_config.ASTEnableDecimalRedrawBar ? "N1" : "N0")
+                                : ""))
+                        : (_config.ASTShowRedrawCooldownTextBar
+                            ? redrawCastInfo.ToString(_config.ASTEnableDecimalRedrawBar ? "N1" : "N0")
+                            : "")) + (_config.ASTShowRedrawTextBar ? " [" + redrawStacks.ToString("N0") + "]" : ""))
+                    : (_config.ASTShowRedrawTextBar ? redrawStacks.ToString("N0") : "");
```
Tischel, line 313: I know nested if statements are lame but this line is really hard to read lol.
Try to clean it up a bit if possible please.

DelvUI/Interface/AstrologianHudWindow.cs
```diff
@@ -368,59 +365,271 @@ internal class AstrologianHudWindow : HudWindow {
...
 
+            ;
+
             var drawList = ImGui.GetWindowDrawList();
             bar.Build().Draw(drawList, PluginConfiguration);
         }
     }
+
+    [Serializable]
+    public class AstrologianHudConfig : PluginConfigObject
```
Tischel, line 442: Awesome!
Really minor, but you could remove the "AST" prefix from the properties since its now redundant with the name of the class it contains them.

Tischel (changes requested):
See my comments above.

DelvUI/Interface/AstrologianHudWindow.cs
Vayhyr: It's was a safety check, since there is an unsafe code bit after that.
I removed it since it's being taken care of before the draw.

DelvUI/Interface/AstrologianHudWindow.cs
```diff
@@ -313,52 +287,133 @@ internal class AstrologianHudWindow : HudWindow {
...
+                }
+                else {
+                    if (_config.ShowRedrawTextBar) {
+                        redrawText = redrawStacks.ToString("N0");
+                    }
+                    else {
+                        redrawText = "";
+                    }
+                }
+
```
Tischel, line 371: Ok this is still a monster haha
I think you should refactor it into a function like this (the code below might have some errors, please double check):

```
protected string RedrawText(int redrawStacks) {
    var ShowRedrawBar = true;
    var redrawCastInfo = 0;
    var EnableRedrawCooldownCumulated = false;
    var ShowRedrawCooldownTextBar = false;
    var EnableDecimalRedrawBar = false;
    var ShowRedrawTextBar = true;

    if (!ShowRedrawBar) {
        return "";
    }

    if (redrawCastInfo <= 0) {
        return ShowRedrawTextBar ? redrawStacks.ToString("N0") : "";
    }

    if (!EnableRedrawCooldownCumulated && (redrawCastInfo % 30) == 0) {
        return "30";
    }

    if (ShowRedrawCooldownTextBar) {
        var format = EnableDecimalRedrawBar ? "N1" : "N0";
        return ShowRedrawTextBar ? (redrawCastInfo % 30).ToString(format) + " [" + redrawStacks.ToString(format) + "]" : (redrawCastInfo % 30).ToString(format);
    }

    return ShowRedrawTextBar ? redrawStacks.ToString("N0") : "";
}
```

Tischel (changes requested):
See comment above about redraw text code.
Also we've recently added a `.editorconfig` file to the project. This should let your editor of choice know the desired style for the code.
You need to make sure the code respects that config.

PR #244, opened 2021-09-03
Improvements on config system

Tischel:
**New classes PluginConfigObject and PluginConfigColor**
Created these 2 with the purpose of removing the enormous color map on PluginConfiguration. The idea is that each element will have it's own config. The config objects will have all the necessary properties for that particular element, along with a function to draw the settings screen for itself.
PluginConfigColor is a little helper class to automatically generate the color maps for a Vector4. It provides easy access to the base, background and gradient colors, along with the full map.

Updated Buff/Debuff and BLM configs to accommodate with these new classes and to serve as an example on how to use them.
Hopefully the rest of the elements get slowly ported over this system over time.

Starting the modularization of this will help in case we want to automatize the settings drawing in the future.

Note: Lots of "fake" changes being flagged due to difference in end line characters :/

### DIFF
--- a/DelvUI/Config/ConfigurationWindow.cs
+++ b/DelvUI/Config/ConfigurationWindow.cs
@@ -4,9 +4,9 @@ using System.Numerics;
 using Dalamud.Interface;
 using ImGuiNET;
 using Dalamud.Plugin;
-using DelvUI.Interface.StatusEffects;
 
-namespace DelvUI.Interface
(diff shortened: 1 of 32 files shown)


Kanubiz (approved):

h4cky:
Awesome!

I'll finish implementing the config system in WarriorHudWindow later today.

PR #253, opened 2021-09-04
[DO NOT MERGE] Complete configuration refactor

### LINKED ISSUE #105: Config Restructure
- [ ] Dynamic Population of the properties
- [ ] Icons on job tabs
- [x] Non job configs
- [x] Job configs
- [x] Structure
- [ ] no resize to inner childs
- [ ] styling everything

Bedo9041:
Closes #105.
This PR aims to completely replace the current config system with a new system that will: 
1. Split the PluginConfiguration code into relevant sections 
2. Split the config file into individual json files on disk 
3. Dynamically draw the configuration window based upon the fields in each configuration object
4. Allow versioning of each config object to better support breaking config changes.
Of these four, currently the first three are implemented in the framework, but the vast majority of the configuration needs to be moved to the new system (only BLM is currently implemented).
Currently there is an issue where in configuration files can be loaded when the plugin is first loaded, however if the plugin is the unloaded and reloaded, Json.NET fails to read the configuration files. While this wouldn't be an issue for users, it would severely hurt developers.
Due to both the incompleteness of the current solution and the aforementioned bug this PR should not be merged and just used for review to see whether there is an agreement on using this approach.

### DIFF
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -1,0 +1,108 @@
+﻿using System;
+using ImGuiNET;
+
+namespace DelvUI.Config.Attributes
+{
+    [AttributeUsage(AttributeTargets.Field)]
+    public class CheckboxAttribute : Attribute
+    {
+        public string friendlyName;
+
+        public CheckboxAttribute(string friendlyName)
+        {
+            this.friendlyName = friendlyName;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Field)]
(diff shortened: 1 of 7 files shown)


Kanubiz (changes requested):
just adding this so i dont accidentally merge, i was about to some time ago before you changed the title

h4cky:
As can be seen above from my ref to this PR i like the idea very much.

Can we make the components like DragFloat(2) to save their value to the json on focusLoss instead of onValueChange, currently its endless overwriting of the json profile for every pixel changed while dragging.

Bedo9041:
Now contained in feature branch of main repo

PR #273, opened 2021-09-05
fix: shield is drawn on top of borders

DomGrieco:
[Shields are drawn on top of the border](https://app.gitkraken.com/glo/view/card/437e6e454c7e48b5b401a5ef97ad7268)


DomGrieco:
Missed something

Issue #245, opened 2021-09-04
Standardize Offset Implementation & Configuration


poly1k:
The same offset values for a bar should mean the same position on the screen, and be consistent across all jobs.

- Jobs should not inherit any offsets from HudWindow.
- Base offsets for all jobs should default to 0/0.
- Zero X/Y offset for any individual bar should mean center of screen.
- Zero offset means the bar is centered, and width/height grows the bar center-out (meaning the offset for any given bar is Center + BarOffset - BarSize / 2f.

poly1k:
![](https://github.trello.services/images/mini-trello-icon.png) [Standardize Offset Implementation & Configuration](https://trello.com/c/6S8XRCtH/42-standardize-offset-implementation-configuration)

poly1k:
Resolved.

PR #276, opened 2021-09-05
Do BRD config, also fixed baseoffset 0, 0)

Zeffuro:
Also did Base Offset to 0, 0 for BRD: #245


Zeffuro:
[BRD - Configuration Migration](https://trello.com/c/W3Aoi3wC/27-brd-configuration-migration)

PR #280, opened 2021-09-05
Fixed a bug where configs were not loaded from file

nilaykumar:
Configs should now be loaded if they exist (instead of resetting to default) in the new configuration framework


jdsmith2816 (changes requested):
just marking as change needed; we chatted about fixing this in discord

PR #282, opened 2021-09-05
[DO NOT MERGE] Hud window refactor

Tischel:
WIP of the HudWindow refactor

Feel free to take a look and give input / feedback.

Some notes:
* Some code I had already done in my party frames branch, like the JobsHelper class or the PropertyChangedEventHandler addition to config objects. Also the shields / overshields drawing methods I already had re-worked before.
* Every unit frame comes from the same class UnitFrameHud which has its own UnitFrameConfig. 
* Labels also are a separated class called LabelHud supported but LabelConfig. This config for now only has Text and Anchor being configurable. Eventually it should have font, font size, colors, etc as well.

Next steps:
* Keep gutting HudWindow up into smaller classes: CastbarHud, MPTickerHud, GCDIndicatorHUD are coming next probably.
* Even though I'm creating config objects for all the elements, I'm not really bothering with drawing the setting windows for them because this will be replacing by the automatic system made by Bedo.

**Note: Refactor is pretty much ready. Just waiting on beta1 to go live so I can work on making this work with the automatic config window system.**


Kanubiz (changes requested):
no merge self reminder

PR #300, opened 2021-09-06
DRG new config minor fixes

asgoel:
Change to use the Origin approach with default base offset set to 0, 0


jdsmith2816:
Need to fix the linting errors

asgoel:
@jdsmith2816 fixed

PR #303, opened 2021-09-07
Added version number display in new config UI

nilaykumar:
Also moved the Donate! button back to the right, where it used to be.
[Demo](https://cdn.discordapp.com/attachments/879308602328903680/884603358676156506/unknown.png)


nilaykumar:
[PluginLog("Plugin Versionnumber") on load](https://trello.com/c/t8FWmd5l/35-pluginlogplugin-versionnumber-on-load)

PR #313, opened 2021-09-07
Blm config improvements

Tischel:
* BLM now almost completely using bar builder (some stuff in mana bar still isnt)
* Every BLM bar has its own offset and size (including procs and dots)
* Fixed bug in bar builder when drain is flipped with multiple inner bars


poly1k:
:shipit:

PR #322, opened 2021-09-08
feat: preview player castbar

DomGrieco:
Show a preview for player castbar when enabled in Config. If config is closed, test castbar will be disabled.


DomGrieco:
[Preview CastBars](https://trello.com/c/4TVMYYY9/36-preview-castbars)

PR #330, opened 2021-09-08
MNK Forms bar fix

TaigaXIV:
enabled by default, should be disabled by default, fixed.


TaigaXIV:
[Fix MNK Hud positions](https://trello.com/c/0PRB7cMp/96-fix-mnk-hud-positions)

PR #332, opened 2021-09-08
[DO NOT MERGE] Hud window refactor merge

Tischel:
This will be the branch where I'm gonna unify the config-refactor stuff with the hud window changes.

There's a decent amount of work that needs to be done to make that happen, most importantly changes on the ConfigNode class and the overall config system to support nested config objects and subclassed config objects.


Kanubiz (changes requested):
no merge reminder

PR #333, opened 2021-09-08
Feat/import export

nilaykumar:
Added functionality to the new config window for importing and export base 64 strings. Let me emphasize that the UI is a WIP -- I will let others more familiar with ImGui make it look good.

The way import/export is done is by adding the functionality to `ConfigPageNode` (the lowest nodes of the config tree). If the import/export GUI is not desired for a given `PluginConfigObject`, simply set `portable = false` in its constructor (see `TankHudConfig`, for example). As a consequence there is not yet a way to export all jobs at once. This could be added later as a new `SelectionNode` under `Job Specific Bars`, which would contain a `ConfigPageNode` (with `portable = false`). But I refrained from doing so, since I'm not sure what the plan is for that.

A quick note on import/export. Because there are currently two different configuration windows, there are two different import/export systems (and reading/writing of configs to file). As the old window gets refactored to the new window, the new import/export should work automatically (given that the refactoring is done correctly).

image

### DIFF
--- a/DelvUI/Config/PluginConfigObject.cs
+++ b/DelvUI/Config/PluginConfigObject.cs
@@ -9,7 +9,6 @@ namespace DelvUI.Config
     [Serializable]
     public abstract class PluginConfigObject
     {
-        public bool portable { get; set; } = true;
 
         protected bool ColorEdit4(string label, ref PluginConfigColor color)
         {
--- a/DelvUI/Config/PluginConfigObject.cs
+++ b/DelvUI/Config/PluginConfigObject.cs
@@ -9,6 +9,8 @@ namespace DelvUI.Config
     [Serializable]
     public abstract class PluginConfigObject
     {
+        public bool portable { get; set; } = true;
+
         protected bool ColorEdit4(string label, ref PluginConfigColor color)
         {
             var vector = color.Vector;
--- a/DelvUI/Interface/NinjaHudWindow.cs
+++ b/DelvUI/Interface/NinjaHudWindow.cs
@@ -339,7 +339,7 @@ namespace DelvUI.Interface
 
         [Checkbox("Show Suiton Bar")]
         [CollapseControl(20, 3)]
-        public bool ShowSuitonBar = true;
+        public bool ShowSuitonBar = false;
 
         [Checkbox("Show Suiton Bar Text")]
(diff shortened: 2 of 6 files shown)


nilaykumar:
[Add import/export string functionality for each appropriate ConfigNode](https://trello.com/c/dVzhlsvz/70-add-import-export-string-functionality-for-each-appropriate-confignode)

DelvUI/Interface/NinjaHudWindow.cs
```diff
@@ -339,7 +339,7 @@ public class NinjaHudConfig : PluginConfigObject
 
         [Checkbox("Show Suiton Bar")]
         [CollapseControl(20, 3)]
-        public bool ShowSuitonBar = true;
+        public bool ShowSuitonBar = false;
```
Bedo9041, line 342: I would personally generally avoid unrelated changes, such as change the default for a particular value. Not necessarily because it's a bad thing, but because such changes can be merged much more quickly if they are separate.

Bedo9041 (approved):
I have no issues with the way that these have been implemented, but I'm unsure the specifics of how this may potentially act with the changes in #332. @Tischel would be able to provide more feedback as to whether this will be entirely incompatible; will just need a revisit; or will be immediately compatible.
Another thing to consider is whether portability should be set in a constructor or as a something like an annotation on the class. I don't really have an opinion either way, but others might.

DelvUI/Interface/NinjaHudWindow.cs
Tischel, line 342: Prob unwanted change

DelvUI/Config/PluginConfigObject.cs
```diff
@@ -9,6 +9,8 @@ namespace DelvUI.Config
     [Serializable]
     public abstract class PluginConfigObject
     {
+        public bool portable { get; set; } = true;
```
Tischel, line 12: Please change to "Portable" to keep the naming convention within the code consistent 🙏

Tischel (changes requested):
See messages above.

nilaykumar:
Thanks for the comments!
I like the idea of portability being an attribute -- I didn't like the constructor approach, but didn't think of attributes. I'll go ahead and implement that.

PR #345, opened 2021-09-09
Add in PvP version of DualCast

Zeffuro:


Zeffuro:
We don't want PvP specific stuff so closing this and we'll have to consider hiding job huds in general in PvP zones.

PR #375, opened 2021-09-10
Hud window refactor

Tischel:
### Big structure change
* **HudWindow** removed. All elements previously drawn in this class were moved to their own files.
* **PluginConfiguration** and **ConfigurationWindow** removed. This classes were replaced by **ConfigurationManager** and **ConfigNode** in beta1. There were some settings still living in them, but with this PR they are no longer needed.
* New classes **HudManager** and **HudElement**. All elements on the screen come from **HudElement** (literally all of them). All **HudElements** contain a config object. **HudManager** has a list of all the elements on the screen and draws them when appropiate.
* The logic for job swap detection was moved from **Plugin** to **HudManager**.
* All job pack classes were moved to a Jobs folder and renamed for consistency (removed the "Window" from the name, for example: **BlackMageHud**)
* All job pack config objects were renamed for consistency (removed the "Hud" from the name, for example: **BlackMageConfig**)

### Config Object Features
* **PluginConfigObject** subclasses now display the attributes from the bases classes as well when drawing the config windows. * Added a _Enabled_ property to **PluginConfigObject**. Since all config objects are of this type, this results in the ability to enable/disable any element individually.
* Added **MovableConfigObject** (subclass of **PluginConfigObject**). This class has a _Position_ and _Size_ property. Most elements config objects are of this type. This means that all config objects that subclass **MovableConfigObject** will have a _Position_ and _Size_ setting by default on their settings (on top of the _Enabled_ one from **PluginConfigObject**).
* Added the ability to have nested config objects. Using the _NestedConfig_ attribute type, a config object and have inside one or multiple config objects. This will result in the attributes of the nested config to be displayed in the settings window automatically. This allows to re-use generic config objects (**LabelConfig** and **EditableLabelConfig** are a good example of that).
* Added _Combo_ attribute type.
* **ConfigurationManager** and **ConfigNode** now use generics to construct the tree (no need to pass "dummy" instances of objects any more).
* **ConfigNode** has an internal map to prevent the re-construction of config pages.

### Other Changes
* The **JobsHelper** class has a lot of methods for job's roles and now contains the job ids.
* Added a **FontsManager** class. Right now it's just a singleton that has the reference to the font. This was done just because **PluginConfiguration**, in the future this class should be expanded to provide font support for the project.
* Added the "fill health first" feature for shields.

### Known Issues
* It is currently impossible to hide attributes from parent config objects types. For example, for **LabelConfig**, displaying a _Size_ setting doesn't make sense. However, right now it's not possible to hide, resulting in unwanted clutter in the config screen. I have a few ideas on how to fix this but haven't put the time into it yet.
* I am not able to test every job 100% so it is pretty likely that I introduced (mostly visual) bugs into some job packs when porting them to the new structure.
* The special "auras" status effect list was removed with this PR. This feature needs to be revisited and we should not include it until we can display a configuration window for it that makes sense (it should let the user add / remove effects they want to track at least).

### Images
image
image
image
image
image
image


DelvUI/Interface/Jobs/BaseJobsConfig.cs
```diff
@@ -0,0 +1,63 @@
+﻿using DelvUI.Helpers;
+using Newtonsoft.Json;
+
+namespace DelvUI.Interface.Jobs
+{
+    public class BaseJobsConfig : JobConfig
+    {
+        public BaseJobsConfig()
+        {
+            UseDefaulyPrimaryResourceBar = true;
```
nilaykumar, line 10: Small typo here

DelvUI/Plugin.cs
```diff
@@ -187,29 +171,17 @@ private void Draw()
                          || _pluginInterface.ClientState.Condition[ConditionFlag.BetweenAreas51];
```
nilaykumar, line 171: Replace `_pluginInterface` with singleton reference

nilaykumar (commented):
Looks great! Thanks for your hard work!

DelvUI/Interface/GeneralElements/CastbarConfig.cs
```diff
@@ -0,0 +1,122 @@
...
+            return new TargetCastbarConfig(pos, size, castNameConfig, castTimeConfig);
+        }
+    }
+
+    [Serializable]
+    public abstract class CastbarConfig : MovablePluginConfigObject
+    {
+        [Checkbox("Preview")]
+        [Order(20)]
+        public bool Preview = false;
```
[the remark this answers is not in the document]
Tischel: Yea I agree. I didn't look much into it, to be completely honest so not sure how hard it'd be. When we get the properties we just get all the properties, at that point im not sure if theres a way to know if it comes from the type you're inspecting or a base type.
This ordering issue is not a problem for nested config objects, luckily.

DelvUI/Plugin.cs
Tischel: No need to do it in the class that has the actual reference.

jdsmith2816:
Going to go ahead and merge this in so we can all get busy testing/fixing/adding but please do still provide feedback when you have time @DomGrieco and @Bedo9041 

This is probably still the cleanest place to initially discuss it even if the PR is already merged.

PR #385, opened 2021-09-11
DragDropHorizontal Attribute and SAM examples

chipget:
Adds a DragDropHorizontal attribute, which controls a list of integers meant to represent the ordering of a set of elements on the UI. The field that the attribute is set on will appear in the config window as a set of (uncolored) labeled buttons that can be clicked and dragged to swap the values in the field. I've included two examples - SAM Shifu/Jinpu ordering and SAM sen ordering.

Note: I tried to implement this with ImGui DragDropSource/DragDropTargets, but found that if we call SetDragDropPayload with any payload size higher than 0, the game immediately crashes as soon as you click one of the buttons. Thus, I opted for this more primitive but stable implementation that just uses MouseDelta values when clicking one of the buttons.

### DIFF
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -155,7 +155,7 @@ namespace DelvUI.Config.Attributes
         public string friendlyName;
         public string[] names;
 
-        public DragDropHorizontalAttribute(string friendlyName, string[] names)
+        public DragDropHorizontalAttribute(string friendlyName, params string[] names)
         {
             this.friendlyName = friendlyName;
             this.names = names;
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -149,6 +149,19 @@ namespace DelvUI.Config.Attributes
         }
     }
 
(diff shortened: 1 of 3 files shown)


jdsmith2816:
@Tischel Can you look at this please?

DelvUI/Config/Attributes/ConfigTypeAttributes.cs
```diff
@@ -149,6 +149,19 @@ public CollapseWithAttribute(int pos, int id)
         }
     }
 
+    [AttributeUsage(AttributeTargets.Field)]
+    public class DragDropHorizontalAttribute : Attribute
+    {   
+        public string friendlyName;
+        public string[] names;
+
+        public DragDropHorizontalAttribute(string friendlyName, string[] names)
```
Tischel, line 158: It'd be great if you can add the _params_ keyword to the _names_ parameter, which allows to use the attribute like thiis:
`[DragDropHorizontal("Friendly Name", "Name1", "Name2"....)]`

You can check the **ComboAttribute** for an example.

Tischel (approved):
Looks good to me, just a minor comment!

DelvUI/Config/Attributes/ConfigTypeAttributes.cs
chipget: Thanks for the suggestion. I've implemented this change now.

Issue #355, opened 2021-09-09
[Suggestion]: 


Drizoon:
### What's your suggestion?

Would it be possible to add text tag for First name only? OR at the very least just First name and first letter of last name.

nilaykumar:
![](https://github.trello.services/images/mini-trello-icon.png) [First name only text tag](https://trello.com/c/oeF5TT0g/114-first-name-only-text-tag)

PR #389, opened 2021-09-11
Added more text-tag options

nilaykumar:
Added:
- `[name:first]`
- `[name:last]`
- `[name:first-initial]`
- `[name:last-initial]`
- `[job]`
as suggested in part by #355.
Also added a dictionary to JobHelper to convert from JobID to the string job name (i.e. JobID.NIN -> "NIN")

The Wiki page https://github.com/DelvUI/DelvUI/wiki/Tags should be updated (with a clear notice that these are available as of v0.2).

image


nilaykumar:
[First name only text tag](https://trello.com/c/oeF5TT0g/114-first-name-only-text-tag)

Tischel:
What's the user for the job names map?

Issue #388, opened 2021-09-11
[Suggestion]: Hide DelvUI outside of combat


h4cky:
### What's your suggestion?

As the title says ability to hide DelvUI outside of combat.

1) Basic
- hide (only) DelvUI Hud's outside of combat and PlayerHud will hide if CurrentHP == MaxHP

2) Extended
- Hides only specific DelvUI elements outside of combat.
- Hide specific Hotbars outside of combat, for example Hotbars nears JobHud used to track CD's
- Hide specific Hotbars when in combat (?)
- Show/Hide DelvUI on specific HUD Layout profiles? e.g I use HUD Layout 1, as my combat profile - shows DelvUI only when I am on this profile, otherwise hides it.

General:
- ability to hide Job Gauges
- ability to hide game castbar (?)

h4cky:
![](https://github.trello.services/images/mini-trello-icon.png) [Feature: Hide DelvUI hud](https://trello.com/c/3PqWjVER/144-feature-hide-delvui-hud)

PR #391, opened 2021-09-11
[DO NOT MERGE]Basic Hide HUD

h4cky:


h4cky:
@Tischel please check it out :)

DelvUI/Interface/HudElement.cs
```diff
@@ -1,12 +1,16 @@
...
 using System.Numerics;
+using DelvUI.Enums;
 
 namespace DelvUI.Interface
 {
     public abstract class HudElement
     {
         protected MovablePluginConfigObject _config;
+
+        public HudType hudType = HudType.None;
```
Tischel, line 12: If you wanna go the enum route, you should make this field _abstract_ and overwrite it in the elements (instead of hardcoding the type for every element in the HudManager).

DelvUI/Interface/GeneralElements/HideHudConfig.cs
```diff
@@ -0,0 +1,33 @@
...
+        /**
+         * MOVE THIS METHOD TO IT'S OWN CLASS
+         **/
+
+        public new static HideHudConfig DefaultConfig()
+        {
+            return new HideHudConfig();
+        }
+    }
+}
```
Tischel, line 33: I'm not sure if we really want a complicated logic for the whole thing. I think initially the idea is just a generic "hide hud out of combat" feature.

However, if you really wanted to have each element configured individually, think you should do it like this:
* All elements could have a "HideConfig" that can be changed to different modes.
* The HudManager / HideHelper would be able to tell for a HideConfig if it should be visible at the moment.

That makes the feature a bit more powerful because it'll result in all elements having an individual setting that the user can change, instead of hardcoding the logic one by one in the code.

Zeffuro: Just in case you weren't aware yet. There's an Github issue and Trello card on this where we can discuss it.
#388 
https://trello.com/c/3PqWjVER/144-feature-hide-delvui-hud-discussion

PR #393, opened 2021-09-11
chore: Add static accessors for commonly used PluginInterface properties

jdsmith2816:
This brings the beta2 and apiv4 branches much closer together and will
result in less maintenance work on every update.

### DIFF
--- a/DelvUI/Interface/Jobs/JobHud.cs
+++ b/DelvUI/Interface/Jobs/JobHud.cs
@@ -6,7 +6,7 @@ namespace DelvUI.Interface.Jobs
 {
     public class JobHud : HudElement, IHudElementWithActor
     {
-        protected DalamudPluginInterface PluginInterface => Plugin.GetPluginInterface();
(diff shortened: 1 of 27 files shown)


DelvUI/Interface/Jobs/JobHud.cs
```diff
@@ -6,7 +6,7 @@ namespace DelvUI.Interface.Jobs
 {
     public class JobHud : HudElement, IHudElementWithActor
     {
-        protected DalamudPluginInterface PluginInterface => Plugin.GetPluginInterface();
+        protected DalamudPluginInterface PluginInterface => Plugin.PluginInterface;
```
Tischel, line 9: This property might not be needed anymore. I think you replaced all the uses already.

jdsmith2816: True, but it's there on the v4 side and is used for.... something.  I forget what.  Figure may as well keep it there to reduce the amount of difference between the two.

PR #394, opened 2021-09-11
Implementing tooltips (used only in status effect lists right now)

Tischel:
image

Had to do some hacky regex and str replace to fix the data.
Some tooltips have tags for color and "glow" that come in very unfriendly unicode characters.
I couldn't find a way to get rid of them "nicely" so the SanitizeText method is a bit ugly.

### DIFF
--- a/DelvUI/Helpers/TooltipsHelper.cs
+++ b/DelvUI/Helpers/TooltipsHelper.cs
@@ -1,0 +1,187 @@
+﻿using Dalamud.Plugin;
+using DelvUI.Config;
+using DelvUI.Config.Attributes;
+using ImGuiNET;
+using System;
+using System.Numerics;
+using System.Text.RegularExpressions;
+
+namespace DelvUI.Helpers
+{
+    public class TooltipsHelper
+    {
+        #region Singleton
+        private TooltipsHelper()
+        {
+            _config = ConfigurationManager.GetInstance().GetConfigObject<TooltipsConfig>();
+        }
+
+        public static void Initialize() { Instance = new TooltipsHelper(); }
+
+        public static TooltipsHelper Instance { get; private set; }
+
+        #endregion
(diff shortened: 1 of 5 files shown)


DelvUI/Helpers/TooltipsHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+        private string _previousRawText = null;
+
+        private Vector2 _position;
+        private Vector2 _size;
+
+        public void ShowTooltipOnCursor(string text, string title = null)
+        {
+            ShowTooltip(text, ImGui.GetMousePos(), title);
+        }
+        public void ShowTooltip(string text, Vector2 position, string title = null)
```
jdsmith2816, line 43:
```suggestion
        }
        
        public void ShowTooltip(string text, Vector2 position, string title = null)
```

Triggered

Tischel: Forgive me father for I have sinned...

DelvUI/Helpers/TooltipsHelper.cs
```diff
@@ -0,0 +1,187 @@
...
+                if (match.Groups.Count > 1)
+                {
+                    result = result.Replace(match.Value, match.Groups[1].Value);
+                }
+            }
+
+            return result;
+        }
+
+        private Vector2 CalculatePosition(Vector2 position, Vector2 size)
```
jdsmith2816, line 147: Maybe ConstrainPosition ?

jdsmith2816 (approved):
LGTM other than a couple minor comments 👍

DelvUI/Helpers/TooltipsHelper.cs
Tischel: sure

PR #399, opened 2021-09-13
Remove buffs with right click

Tischel:
A few notes:
* The feature uses the **/statusoff** command to remove buffs. This means we are sending a chat message in place of the user to achieve this.
* For this reason, I've created a **ChatHelper** class that allows us to send ANY message as if it was sent by the player. Right now this is only used for this particular feature, but it could be used for other things. However, we should be very careful on the things we do with this. I think we should limit its use as much as possible and only do it for small stuff like this.
* **StatusEffectListHud** has now its own window to be able to receive inputs (visually no changes though).
* Added some methods in **DrawHelper** to draw text and icons by using a pre-determined imgui draw list.
* Added a DrawOutlineText method that receives thickness as parameter (the duration and stacks on some status effects were hard to read, by increasing the outline size it looks better)

image

### DIFF
--- a/DelvUI/Helpers/ChatHelper.cs
+++ b/DelvUI/Helpers/ChatHelper.cs
@@ -9,15 +9,13 @@ namespace DelvUI.Helpers
         #region Singleton
         private ChatHelper()
         {
-            var scanner = Plugin.PluginInterface.TargetModuleScanner;
-
-            IntPtr baseUi = scanner.GetStaticAddressFromSig("48 8B 0D ?? ?? ?? ?? 48 8D 54 24 ?? 48 83 C1 10 E8");
-            IntPtr uiModule = scanner.ScanText("E8 ?? ?? ?? ?? 48 83 7F ?? 00 48 8B F0");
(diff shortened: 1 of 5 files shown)


Zeffuro (commented):
All looks good to me

DelvUI/Helpers/ChatHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Text;
+
+namespace DelvUI.Helpers
+{
+    public class ChatHelper
+    {
+        #region Singleton
+        private ChatHelper()
+        {
+            var scanner = Plugin.PluginInterface.TargetModuleScanner;
```
jdsmith2816, line 12: Everything we need to use from PluginInterface is now attached to Plugin directly;  could you use those instead?  It'll end up being less differences on the apiv4 branch.

```suggestion
            var scanner = Plugin.SigScanner;
```

Tischel: sure

Tischel:
I just noticed separating the status effect lists into a window, makes them be drawn on top of tooltips.
I'll fix this in a another PR (gotta make tooltips have their own window, basically)

Issue #377, opened 2021-09-10
[Suggestion]: A 0.5 second indicator on the GCD timer bar


Clickeddaisy:
### What's your suggestion?

Like the title suggests, On the GCD timer bar i would like an option where the last 0.5 seconds to have an indicator so you know exactly when you can queue up your next GCD ability or change the colour of the whole bar when it has 0.5 seconds and less left.

And just as a side request for the GCD timer instead of it going away when it is finished have it as full instead of empty when it finishes.

Love the Plugin BTW keep up the good work :)

Laxxzz:
![](https://github.trello.services/images/mini-trello-icon.png) [Add indicator on GCD bar To Show When You can que Next Ability](https://trello.com/c/lo3Y4nHd/139-add-indicator-on-gcd-bar-to-show-when-you-can-que-next-ability)

Tischel:
> And just as a side request for the GCD timer instead of it going away when it is finished have it as full instead of empty when it finishes.

This is already possible with the "Always Show" setting.

Clickeddaisy:
> 
> 
> > And just as a side request for the GCD timer instead of it going away when it is finished have it as full instead of empty when it finishes.
> 
> This is already possible with the "Always Show" setting.

Sort of. It will always show the background. What i want is when the bar fills it stays filled instead of going away.

PR #400, opened 2021-09-13
Added a queue window indication on the gcd bar

chipget:
This PR implements issue #377 by partitioning the gcd bar into two chunks. The first chunk corresponds to the first (GCD - 0.5) seconds and the second chunk corresponds to the latter 0.5 seconds. The indicator is implemented as padding between these two chunks, which has an adjustable size in the gcd configuration window.

### DIFF
--- a/DelvUI/Interface/GeneralElements/GCDIndicatorConfig.cs
+++ b/DelvUI/Interface/GeneralElements/GCDIndicatorConfig.cs
@@ -22,6 +22,10 @@ namespace DelvUI.Interface.GeneralElements
         [Order(30)]
         public bool VerticalMode = false;
 
+        [DragFloat("GCD Queue Padding Size")]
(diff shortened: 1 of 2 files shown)


Tischel:
You should have it as something that can be enabled or disabled. With this code it'll always have chunks.
Also it'd be nice if you could show a gif or video to see how it looks.

Tischel (changes requested):
You should have it as something that can be enabled or disabled. With this code it'll always have chunks.
Also it'd be nice if you could show a gif or video to see how it looks.

chipget:
Actually, while it looks pretty good in vertical mode, it looks fairly bad in horizontal mode. I'm going to close this and change it to color indicators.

gcdqueuehorizontal
gcdqueuevertical
gcdqueuewideractive

PR #416, opened 2021-09-15
Draggable elements

Tischel:
Mostly what's seen here plus some bug fixes and the ability to see the selected element's anchor point when dragging (see picture below):

https://www.youtube.com/watch?v=xoclRzAASYs

image

**NOTE: The default position for all job packs have been changed. Having the base offset at (0, 0) makes no sense with this system where you can visually see the position of each element in relation to the origin (center of the screen). For this reason all the job packs base offset and their sub elements positions have been adjusted.**


Tischel:
Still work needed on almost every job pack, please don't merge until I'm done with that.

Tischel:
Should be good to merge now.

PR #417, opened 2021-09-16
fix: fix label anchors and draw labels relative to their parent HudElement

lichie567:
This change allows labels on unit frames and cast bars to be drawn with knowledge of the size of the parent element. This way, the text labels will stay anchored and move while the parent element is resized.

### DIFF
--- a/DelvUI/Interface/GeneralElements/CastbarConfig.cs
+++ b/DelvUI/Interface/GeneralElements/CastbarConfig.cs
@@ -37,8 +37,8 @@ namespace DelvUI.Interface.GeneralElements
             var size = new Vector2(254, 24);
             var pos = new Vector2(0, HUDConstants.PlayerCastbarY);
 
-            var castNameConfig = new LabelConfig(new Vector2(-size.X / 2f + size.Y + 5, 0), "", LabelTextAnchor.Left);
-            var castTimeConfig = new LabelConfig(new Vector2(size.X / 2f - 5, 0), "", LabelTextAnchor.Right);
+            var castNameConfig = new LabelConfig(new Vector2(5, 0), "", LabelTextAnchor.Left);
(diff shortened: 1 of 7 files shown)


Tischel:
I think if we're gonna have something like this the wording needs to change and you need 2 settings. 
Right now you're using the same anchor setting to define the label's anchor point, and the position in which the label will be constrained to the parent.
With this implementation you can't have a label with an anchor point on the left, while being attached to the right side of an element, for example.

Here's how Weakauras does it, and how I think we should have it if we wanna go this route (the label anchor and the relative position to its parent are completely separated):
image

Here's how ElvUI does it (basically same implementation as this PR):
image

What I don't like about it is that for some things you still need to touch the offset, whereas with the way weakaura's does it you could anchor the label to any position inside or outside the parent's frame, without changing the offset.

I do think this change is an improvement but I'd prefer to have the complete system and the label config being the one that decides if the text is drawn relative to the parent's frame or not (since right now it's basically hardcoded in each hud, so even though you still have the old functionality by calling LabelHud.Draw, there's no way to actually use it right now).

jdsmith2816 (changes requested):
marking as changes needed due to tischels comment

lichie567:
I added a new label config setting to allow control of both where the label anchors to the frame, and where the text anchors relative to the frame anchor.

Tischel:
> I added a new label config setting to allow control of both where the label anchors to the frame, and where the text anchors relative to the frame anchor.

Awesome!

Tischel:
Should be good to merge once the conflicts are solved (I changed the draw text calls inside label hud today to use the ones that take a draw list as parameter).

lichie567:
Took care of the merge conflicts... FYI this change will almost certainly make existing label configs look crazy, so I recommend resetting to default or just changing your label config to match the defaults I added.

PR #421, opened 2021-09-16
Feat: Hide Hud

h4cky:
This PR implements parts of  [Feature: Hide DelvUI Hud (Discussion)](https://trello.com/c/3PqWjVER/144-feature-hide-delvui-hud-discussion) and some more:

1) Hide HUD options (DelvUI > Misc > Hide Options)
- Hide DelvUI when outside of combat
- Hide Default Job Gauge
- Hide Default Castbar
- Show selected Hotbars only in combat

2) Config property observation
Introduces the ability to track Config changes without checking to them every frame.
- add `public interface IOnChangeEventArgs`
- add `public class OnChangeEventArgs<T> : OnChangeBaseArgs`
- add `public class DynamicList : ConfigAttribute`

Property observation can be seen implemented in `HudHelper` which subscribes for the events sent by `HideHudConfig`.
The chain for the received events is as follows:
ConfigNode (value changed) >> calls the `Config.onValueChangedRegisterEvent(EventArgs)` >> `Config.onValueChangedRegisterEvent(EventArgs)` sends msg to all subscribed classes via `EventHandler<OnChangeBaseArgs> onValueChanged`

### DIFF
--- a/DelvUI/Interface/GeneralElements/HideHudConfig.cs
+++ b/DelvUI/Interface/GeneralElements/HideHudConfig.cs
@@ -3,13 +3,14 @@ using DelvUI.Config;
 using DelvUI.Config.Attributes;
 using Newtonsoft.Json;
 using System.Collections.Generic;
+using System.Numerics;
 
 namespace DelvUI.Interface.GeneralElements
 {
     [Serializable]
     [Section("Misc")]
     [SubSection("Hide Options", 0)]
-    public class HideHudConfig : PluginConfigObject, IOnChangeEventArgs
+    public class HideHudConfig : PluginConfigObject
     {
         [Checkbox("Hide DelvUI outside of combat")]
         [Order(5)]
@@ -32,18 +33,5 @@ namespace DelvUI.Interface.GeneralElements
         public List<string> CombatActionBars = new();
 
         public new static HideHudConfig DefaultConfig() { return new HideHudConfig(); }
-
-        #region OnChangeEventInterface
-
-        // sending event outside of the config
-        public event EventHandler<OnChangeBaseArgs> onValueChanged;
-
-        // received events from the node
-        public void onValueChangedRegisterEvent(OnChangeBaseArgs e)
-        {
-            onValueChanged?.Invoke(this, e);
-        }
-
-        #endregion
     }
 }
--- a/DelvUI/Interface/GeneralElements/HideHudConfig.cs
+++ b/DelvUI/Interface/GeneralElements/HideHudConfig.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using DelvUI.Config;
+using DelvUI.Config.Attributes;
+using Newtonsoft.Json;
+using System.Collections.Generic;
+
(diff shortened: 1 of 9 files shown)


DelvUI/Interface/GeneralElements/HideHudConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+        // sending event outside of the config
+        public event EventHandler<OnChangeBaseArgs> onValueChanged;
+
+        // received events from the node
+        public void onValueChangedRegisterEvent(OnChangeBaseArgs e)
+        {
+            onValueChanged?.Invoke(this, e);
+        }
+
+        #endregion
```
Tischel, line 47: I think you should put this directly on PluginConfigObject so all config objects can benefit from this feature without having to do anything or repeat code.

DelvUI/Interface/HudHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+                {
+                    return false;
+                }
+
+                if (element.ID == HudManager.playerUnitFrameID)
+                {
+                    PlayerCharacter player = Plugin.ClientState.LocalPlayer;
+                    Debug.Assert(player != null, "HudHelper.LocalPlayer is NULL.");
+                    isHidden = isHidden && player.CurrentHp == player.MaxHp;
+                }
```
Tischel, line 124: I think you should either check directly for the type or add a json ignored property on the config objects.
The way you have it right now is not very flexible / scalable.

DelvUI/Interface/HudManager.cs
```diff
@@ -31,6 +32,12 @@ public class HudManager
         private Dictionary<uint, JobHudTypes> _jobsMap;
         private Dictionary<uint, Type> _unsupportedJobsMap;
 
+        public HudHelper helper { get; } = new HudHelper();
```
Tischel, line 35: Any reason this is not private?
Also name should be Helper (if it remains public) or _helper (if its changed to private); to go in line with the current naming conventions in the code.

Tischel (changes requested):
See comments below.

DelvUI/Interface/HudHelper.cs
h4cky: for the castbar thats easy, but for the player UnitFrameHud there are few elements of that type, unless i miss something, or maybe i can check the UnitFrameHud.actor ?

DelvUI/Interface/HudManager.cs
h4cky: sry, that used to be private property, i’ll have it renamed later today

h4cky: I’ll go with public since there is  at least  one case where i need to set flags  from Plugin.Draw

DelvUI/Interface/HudHelper.cs
Tischel: Theres a **PlayerUnitFrameConfig** class.

PR #425, opened 2021-09-17
Configuration Restyling

Kanubiz:
Made a lot of initial changes to the configs overall
Currently restyling the shell
Homogenizing certain colors (some colors were exactly the same color/purpose but in a slightly different shade for example MP bars had like 4 different shades of the same blue)
Changed the gradient to top to bottom
Changed some colors to match MaterialUI

IMPORTANT: Need someone to run a code formatting just in case mine is still broken

### DIFF
--- a/DelvUI/Config/Tree/ConfigNode.cs
+++ b/DelvUI/Config/Tree/ConfigNode.cs
@@ -92,9 +92,11 @@ namespace DelvUI.Config.Tree
             }
 
             var configPageNode = GetOrAddConfig<T>();
+
             if (configPageNode != null && configPageNode.ConfigObject != null)
             {
                 configPageNodesMap.Add(type, configPageNode);
+
                 return (T)configPageNode.ConfigObject;
             }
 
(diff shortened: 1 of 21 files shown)


DelvUI/Config/Tree/ConfigNode.cs
```diff
@@ -146,19 +148,83 @@ public override void LoadBase64String(string[] importStrings)
...
+            ImGui.PopStyleColor();
+            
+            ImGui.PopStyleColor();
+            ImGui.PopStyleColor();
+            
+            ImGui.PopStyleVar();
+            ImGui.PopStyleVar();
+            ImGui.PopStyleVar();
+            ImGui.PopStyleVar();
+            ImGui.PopStyleVar();
```
Tischel, line 204: You can just do **PopStyleColor(14)** and **PopStylevar(5)**

DelvUI/Helpers/TextTags.cs
```diff
@@ -1,100 +1,99 @@
...
+        }
+
+        public static string GenerateFormattedTextFromTags(dynamic actor, string text)
+        {
+            var matches = Regex.Matches(text, @"\[(.*?)\]");
+
+            return matches.Cast<Match>().Aggregate(text, (current, m) => current.Replace(m.Value, ReplaceTagWithString(m.Value, actor)));
+        }
+    }
+}
```
Tischel, line 99: Is there any real change in this file?

Tischel (changes requested):
See comments below.

Also we should probably stop using the color maps and just straight up use PluginConfigColor and its properties. A small refactor on BarBuilder would be needed. Doesn't need to happen on this PR, but hardcoding the color map keys everywhere is not ideal.

DelvUI/Helpers/TextTags.cs
Kanubiz: nop

PR #433, opened 2021-09-18
feat: add configurable anchoring for unitframes, castbars, and misc bars

lichie567:
Added the ability to specify an anchor point for various HudElements:
- UnitFrames
- CastBars
- Misc bars (GCD Indicator, MP Ticker, Primary Resource Bar)

I would like to add this functionality for the job huds at some point, but the current implementation for job huds does not have a pre-configured size which is needed for calculating the offset for the anchoring.


Tischel:
I feel like there's not much to gain by having this but at the same time there's no harm in having it.

If we're putting this in we should at least:
* Create an "Anchor" attribute type so the combo options are not repeated / hardcoded in so many places:
`    [AttributeUsage(AttributeTargets.Field)]
    public class AnchorAttribute : ComboAttribute
    {
        public AnchorAttribute(string friendlyName)
            : base(friendlyName, new string[] { "Center", "Left", "Right", "Top", "TopLeft", "TopRight", "Bottom", "BottomLeft", "BottomRight" })
        {
        }
    }`

* Make sure the anchors are displayed properly when on grid mode. I don't think you can do this in a nice way unless DrawAnchor was part of MovablePluginObject, which is also gonna cause a lot of problems. So not sure what the best way to achieve this is. I guess you could use reflection to check if the object has a DrawAnchor, but that sounds messy and complicated.

Regarding the job packs, ideally each bar would have its individual config. That'd be what makes the most sense with the current system. However I'm not sure we want that, or if we want it right now. Not even mentioning achieving this would be a massive undertaking,

Tischel:
I've been thinking more about this kind of feature.
I think the proper way to approach it would be to actually add a "parent" concent to hud elements.
The position of the element will then be calculated depending on its anchor and the parent position / sizes.

Basically a pretty normal hierarchy based drawing system. We're currently kinda faking it, but if we want features like this it might be worth to implement it properly.

lichie567:
I definitely agree on implementing a real parent/child hierarchy but it would be a pretty big refactor and is probably better suited for beta3. I've got this working in a pretty generic way for now, let me know what you think.

PR #435, opened 2021-09-18
Added separator as a prop

Kanubiz:
PRing now to PR the ninja changes and start working on restyling job packs

### DIFF
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -17,7 +17,6 @@ namespace DelvUI.Config.Attributes
         public CheckboxAttribute(string friendlyName)
         {
             this.friendlyName = friendlyName;
-            separator = false;
         }
     }
 
@@ -36,7 +35,6 @@ namespace DelvUI.Config.Attributes
             min = 1f;
             max = 1000f;
             velocity = 1f;
-            separator = false;
 
         }
     }
@@ -55,7 +53,6 @@ namespace DelvUI.Config.Attributes
             min = 1;
             max = 1000;
             velocity = 1;
-            separator = false;
         }
     }
 
@@ -73,7 +70,6 @@ namespace DelvUI.Config.Attributes
             min = 1f;
             max = 1000f;
             velocity = 1f;
-            separator = false;
         }
     }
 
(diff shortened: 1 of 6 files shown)


DelvUI/Config/Attributes/ConfigTypeAttributes.cs
```diff
@@ -6,14 +6,19 @@ namespace DelvUI.Config.Attributes
...
     [AttributeUsage(AttributeTargets.Field)]
     public class CheckboxAttribute : ConfigAttribute
     {
         public string friendlyName;
 
-        public CheckboxAttribute(string friendlyName) { this.friendlyName = friendlyName; }
+        public CheckboxAttribute(string friendlyName)
+        {
+            this.friendlyName = friendlyName;
+            separator = false;
```
Tischel, line 20: You don't need to do this in every constructor when the base class is already setting the initial value to false

DelvUI/Config/Tree/ConfigNode.cs
```diff
@@ -1038,6 +1035,30 @@ public CategoryField(FieldInfo mainField, PluginConfigObject configObject, strin
...
+                else if (attribute is DragFloatAttribute { separator: true }) {}
+                else if (attribute is DragIntAttribute { separator: true }) {}
+                else if (attribute is DragFloat2Attribute { separator: true }) {}
+                else if (attribute is DragInt2Attribute { separator: true }) {}
+                else if (attribute is InputTextAttribute { separator: true }) {}
+                else if (attribute is ColorEdit4Attribute { separator: true }) {}
+                else if (attribute is ComboAttribute { separator: true }) {}
+                else if (attribute is DragDropHorizontalAttribute { separator: true }) {}
+                else if (attribute is DynamicList { separator: true }) { }
+                else { continue; }
```
Tischel, line 1052: If the only special case is CheckboxAttribute you should do something like this:
```
if (attribute is CheckboxAttribute { friendlyName: "Enabled", separator: true } && ID != null) { }
else if (attribute is ConfigAttribute { separator: true }) {}
else { continue; }
```
Since the property is on the base class thats all you need to check for.

Tischel (changes requested):
See comments below

DelvUI/Config/Tree/ConfigNode.cs
Kanubiz: hmm ok nice, i was struggling to figure this solution

PR #439, opened 2021-09-18
Hide HUD changes, i hope they work.

h4cky:
- Refactor/clean up of HudHelper
- HudHelper is not private property of HudManager
- addressed some cases where in-game JobHud was shown.
- added "I Dont Care About The Sounds" option in settings under "Hide Default Job Gauges"; in that case a lot less traversal of the addonList structs is done.

### DIFF
--- a/DelvUI/Interface/GeneralElements/HideHudConfig.cs
+++ b/DelvUI/Interface/GeneralElements/HideHudConfig.cs
@@ -21,19 +21,23 @@ namespace DelvUI.Interface.GeneralElements
         public bool HideOnlyJobPackHudOutsideOfCombat = false;
 
         [Checkbox("Hide Default Job Gauges", isMonitored = true)]
-        [Order(15)]
+        [CollapseControl(15, 0)]
         public bool HideDefaultJobGauges = false;
+
+        [Checkbox("I Dont Care About The Sound", isMonitored = true)]
+        [CollapseWith(0, 0)]
+        public bool IDontCareAboutTheSounds = false;
 
         [Checkbox("Hide Default Castbar", isMonitored = true)]
         [Order(20)]
         public bool HideDefaultCastbar = false;
 
         [Checkbox("Enable Combat Hotbars", isMonitored = true)]
-        [CollapseControl(25, 0)]
+        [CollapseControl(25, 1)]
         public bool EnableCombatActionBars = false;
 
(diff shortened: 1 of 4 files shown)


DelvUI/Interface/GeneralElements/HideHudConfig.cs
```diff
@@ -21,19 +21,23 @@ public class HideHudConfig : PluginConfigObject
         public bool HideOnlyJobPackHudOutsideOfCombat = false;
 
         [Checkbox("Hide Default Job Gauges", isMonitored = true)]
-        [Order(15)]
+        [CollapseControl(15, 0)]
         public bool HideDefaultJobGauges = false;
 
+        [Checkbox("I Dont Care About The Sound", isMonitored = true)]
+        [CollapseWith(0, 0)]
+        public bool IDontCareAboutTheSounds = false;
```
Tischel, line 29: The naming here is werid.
If disabling this boolean makes the sounds from default job bars not work at all, then change the name to something more intuitive like "Disable Job Gauges Sounds".
If disabling this boolean doesn't ensure anything, then I think it should ensure something haha.

DelvUI/Interface/HudManager.cs
```diff
@@ -283,9 +285,18 @@ protected unsafe bool ShouldBeVisible()
...
             if (!ConfigurationManager.GetInstance().ShowHUD || Plugin.ClientState.LocalPlayer == null)
             {
+                _helper = null;
                 return false;
             }
 
+            if (_helper == null)
+            {
+                _helper = new HudHelper();
+                _helper.Configure(true, true);
```
Tischel, line 295: Whats the reason behind having to re-instantiate the helper every time the hud is hidden or the player reference is lost?

DelvUI/Interface/GeneralElements/HideHudConfig.cs
h4cky: yeah it removes the sound by removing the NodeList from the Job Hud.

DelvUI/Interface/HudManager.cs
h4cky: Logging out of the game and back in with the same character, leads to loss of state in a sense, during first login I was configuring the HudHelper in UpdateJobs() if you properly load a job.

However on subsequent logins with the same character the current job hud is already loaded, HudManager - was never released and so UpdateJobs() method exits at the first If-statement, and that leads to for example combat hotbars showing up, or the ingame job gauge not being hidden, etc.

DelvUI/Interface/GeneralElements/HideHudConfig.cs
Tischel: OK then let's use a better name like "Disable Job Gauge Sounds"

PR #443, opened 2021-09-19
Job Colors for Labels and Primary Resource bar

vond82:
Personal preference to have the job coloring on the text rather than the bars themselves so labels now have the option to show text job colored, disabled by default. Same for the primary resource bar.


vond82:
Realized I should make sure NPCFriendly/Neutral/Hostile etc is used when applicable so converting to draft

vond82:
OK should be ready for review now

PR #445, opened 2021-09-19
Job Colors for Labels and Primary Resource bar

vond82:
- Adds a Job Color option for labels and the primary resourcebar. Disabled by default.
- Uses NPCFriendly/Hostile/Neutral coloring for NPC targets.
- The FriendlyNPC unitframe now updates its color when UseCustomColor is checked.

### DIFF
--- a/DelvUI/Interface/GeneralElements/PrimaryResourceHud.cs
+++ b/DelvUI/Interface/GeneralElements/PrimaryResourceHud.cs
@@ -44,7 +44,22 @@ namespace DelvUI.Interface.GeneralElements
             var drawList = ImGui.GetWindowDrawList();
             drawList.AddRectFilled(startPos, startPos + Config.Size, 0x88000000);
 
-            drawList.AddRectFilledMultiColor(
+            if (Config.UseJobColor)
+            {
+                var color = GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id);
+
+                drawList.AddRectFilledMultiColor(
+                startPos,
+                startPos + new Vector2(Math.Max(1, Config.Size.X * scale), Config.Size.Y),
+                color.TopGradient,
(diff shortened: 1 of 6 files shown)


DelvUI/Interface/GeneralElements/PrimaryResourceHud.cs
```diff
@@ -44,14 +44,30 @@ public override void DrawChildren(Vector2 origin)
...
+            {
+                var color = GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id);
+
+                drawList.AddRectFilledMultiColor(
+                startPos,
+                startPos + new Vector2(Math.Max(1, Config.Size.X * scale), Config.Size.Y),
+                color.TopGradient,
+                color.TopGradient,
+                color.BottomGradient,
+                color.BottomGradient
```
Tischel, line 57: I missed changing these calls to use the DrawHelper.DrawGradientFilledRect earlier, can you replace them with that? So it uses the global gradient setting

Tischel: Also instead of having the same code twice, define the color first, then make the call, like you did in LabelHud

Tischel (changes requested):
Comments below

DelvUI/Interface/GeneralElements/PrimaryResourceHud.cs
vond82: Sure can sort it when daughter is asleep later today

vond82: Ok sorted, also added the call to DrawHelper.DrawGradientFilledRect for the FriendlyNPC unitframe as well since I noticed it was missing there too

PR #452, opened 2021-09-20
Adding a "tracked buffs" list

Tischel:
It has a pre-populated white list of common raid buffs

Also:
* Fixed / re-did the "centered" growth direction option for status effects lists
* Fixed blacklist not being able to have 2 statuses with the same name but different id, while still maintaining the list ordered by name


Kanubiz (changes requested):
Tischel adding target chara to track target debuffs too

PR #456, opened 2021-09-21
adds ability to hide your DelvUI in gold saucer

h4cky:
This PR implements the ability to hide DelvUI in certain ClientState.TerritoryType's which are part of The Gold Saucer. We do no toggling of hotbars so if the user has his Petbar only as player hotbar1 and if the respective bar is hidden it wont be visible. 

Some TerritoryType id might be missing from the list

### DIFF
--- a/DelvUI/Interface/HudHelper.cs
+++ b/DelvUI/Interface/HudHelper.cs
@@ -106,9 +106,7 @@ namespace DelvUI.Interface
             }
 
             // hide in gold saucer
-            var results = GoldSaucerIDs.Where(id => id == Plugin.ClientState.TerritoryType);
-            bool hasZoneIdMatch = results.Count() > 0;
-            if (Config.HideInGoldSaucer && hasZoneIdMatch)
+            if (Config.HideInGoldSaucer && GoldSaucerIDs.Where(id => id == Plugin.ClientState.TerritoryType).Count() > 0)
             {
                 return true;
             }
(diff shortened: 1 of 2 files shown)


DelvUI/Interface/HudHelper.cs
```diff
@@ -104,6 +105,14 @@ public bool IsElementHidden(HudElement element)
...
             }
 
+            // hide in gold saucer
+            var results = GoldSaucerIDs.Where(id => id == Plugin.ClientState.TerritoryType);
+            bool hasZoneIdMatch = results.Count() > 0;
+            if (Config.HideInGoldSaucer && hasZoneIdMatch)
+            {
+                return true;
+            }
+
```
Tischel, line 115: Kinda nitpicking but you should only check the array if the config is enabled, not before.

Tischel (changes requested):

DelvUI/Interface/HudHelper.cs
Tischel: You can just do if (enabled && array.Where(blabla).Count > 0)

conditions get checked in order so if the first one fails the 2nd one shouldn't be executed

h4cky: well at first i tried:
if (Config.HideInGoldSaucer && GoldSaucerIDs.Where(id => id == Plugin.ClientState.TerritoryType).Count() > 0)

and the compiler complained about something so i changed it to the way you see above

PR #467, opened 2021-09-22
a few more variants on the [name:first] texttag

vond82:
3 more variants on the [name:first] texttag:

- [name:first-npcmedium] works the same as [name:first] for players but NPCs/objects will have their names trunctuated to 15 characters.
- [name:first-npclong] works the same as [name:first] for players but NPCs/objects will have their names trunctuated to 20 characters.
- [name:first-npcfull] works the same as [name:first] for players but NPCs/Objects will show their full name.

Just something that came from personal annoyance with seeing targets names cut off, when all I wanted was to remove lastnames from myself and other players


jdsmith2816 (changes requested):
@vond82 will merge if you open a PR to the apiv4 branch as well ;)

PR #472, opened 2021-09-22
use job color as background color & color health bar by health value

lemonlust:
Added the option to color unit frames based on health value. We have the option to choose two colors to `interpolate` between depending on the health value of the unit. I also added the option to color the background by job color as well. The colors are converted from [`sRGB`](https://en.wikipedia.org/wiki/SRGB) color space to [`CEILAB`](https://en.wikipedia.org/wiki/CIELAB_color_space) using [`The Colorful Library`](https://github.com/tompazourek/Colourful) library. Then the colors are interpolated based on the health value of the unit then converted back to [`sRGB`](https://en.wikipedia.org/wiki/SRGB) color space.

I also added the option to display the Full Health or Low Health color when your health value is above or below certain thresholds. By default it will always display the Low Health Color below `25%` before interpolating into the Full Health Color. Full health color will always display above `75%` before interpolating into the Low Health Color.

[`The Colorful Library`](https://github.com/tompazourek/Colourful) can also be helpful for `interpolating` colors for different frames such as color by remaining duration of a buff or color the cast bar based on cast time.

[`The Colorful Library`](https://github.com/tompazourek/Colourful) is a `Open source` .NET library for working with `color spaces`.

The library is written in `C#` and released with an [`MIT license`](https://github.com/tompazourek/Colourful/blob/master/LICENSE)

image

When interpolating from `red` to `green` in `RGB` space, we get this un-intuitive brown in the middle. Furthermore, `lchuv` and `lchab` looks very similar to `xyY` and `xyz`, but changing the initial colors a bit reveals their dissimilarity.

image

`LCHuv` and `LCHab` is very similar to `HSV` and `HSL`, but they do avoid the luminance spikes at yellow and light blue
image


DelvUI/Interface/GeneralElements/UnitFrameHud.cs
```diff
@@ -225,6 +225,11 @@ private uint BackgroundColor(Chara chara)
                 return Config.CustomBackgroundColor.Base;
             }
 
+            if (Config.UseJobColorAsBackgroundColor)
+            {
+                return ImGui.ColorConvertFloat4ToU32(GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id).Vector);
```
Tischel, line 230: `return GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id).Base`

Tischel (changes requested):

DelvUI/Interface/GeneralElements/UnitFrameHud.cs
lemonlust: okay ill make the change here

DelvUI/Interface/GeneralElements/UnitFrameHud.cs
```diff
@@ -222,7 +222,17 @@ private uint BackgroundColor(Chara chara)
...
+                if (Config.UseJobColorAsBackgroundColor)
+                {
+                    color = GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id).Base;
+                    return color;
+                }
+                else
+                {
+                    color = Config.CustomBackgroundColor.Base;
+                }
+                return color;
```
Tischel, line 235: You don't need the variable:
```
if (Config.UseJobColorAsBackgroundColor)
{
    return GlobalColors.Instance.SafeColorForJobId(chara.ClassJob.Id).Base;
}
else
{
    return Config.CustomBackgroundColor.Base;
}
```

Tischel (changes requested):

DelvUI/Interface/GeneralElements/UnitFrameHud.cs
demmy: ternary operator be better

Tischel: sure

lemonlust: made the changes

lemonlust:
image
Health under 30% is red and Health Above 80% is Green

lemonlust (commented):
@Tischel Alright just added a new feature to change frame color by health value

lemonlust (commented):
Convert the RGB values of the two colors to HSL before we interpolate. 
Added config health% thresholds for low/full health colors.

Tischel:
This looks fine to me, not sure if we wanna add it now or wait for beta3.
Also it adds a new dependency (which is cool since it greatly reduces the code), and I don't know what's the deal with that in de apiv4 branch (do we still need ILMerge? can we add dependencies freely?)

@jdsmith2816

jdsmith2816:
@lemonlust github autoclosed these PRs when I merged beta2;  please reopen these changes to beta3 branch with any relevant updates and we can get them merged in

PR #474, opened 2021-09-22
Remove options from collapse and add spacing

Zeffuro:
Removes the options from the circular mode collapse since they're usable outside of circular mode + add spacing.


Zeffuro:
[Circular GCD Indicator](https://trello.com/c/vzLOYjSE/171-circular-gcd-indicator)

PR #475, opened 2021-09-22
Only Show When Active-options

vond82:
This is an attempt for a way to make UIs that are a lot less cluttered with empty bars when out of combat especially, while also being able to pick and choose what you'd like to see meaning you can keep it exactly as it has been.

I've also made some changes to the way text on job bars is displayed by hiding them when at 0 and given the option to hide them completely where missing. 

Only SAM and MNK done so far but I plan to go through all the jobs as best I can.

https://youtu.be/Z5u2DDV3ER8 to showcase why I think this is a lot better personally

### DIFF
--- a/DelvUI/Interface/Jobs/SamuraiHud.cs
+++ b/DelvUI/Interface/Jobs/SamuraiHud.cs
@@ -90,6 +90,11 @@ namespace DelvUI.Interface.Jobs
                 origin.Y + Config.Position.Y + Config.KenkiBarPosition.Y - Config.KenkiBarSize.Y / 2f
             );
 
+            if (gauge.Kenki == 0 && Config.OnlyShowKenkiWhenActive)
+            {
+                return;
+            }
+
             var kenkiBuilder = BarBuilder.Create(pos, Config.KenkiBarSize)
                 .SetBackgroundColor(EmptyColor.Base)
                 .AddInnerBar(gauge.Kenki, 100, Config.KenkiColor);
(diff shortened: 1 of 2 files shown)


vond82:
Just wanting to know if this is something others would like for all jobs or if i should scrap the idea

DelvUI/Interface/Jobs/SamuraiHud.cs
```diff
@@ -239,15 +254,15 @@ public class SamuraiConfig : JobConfig
         [CollapseControl(30, 0)]
         public bool ShowKenkiBar = true;
 
-        [DragFloat2("Kenki Bar Size", max = 2000f)]
+        [DragFloat2("Kenki Bar Size" + "##Kenki", max = 2000f)]
```
h4cky, line 257: For me that, appending "##Kenki", would make sense if there was multiple fields named "Kenki Bar Size" and CategoryField.ID == null (which sadly currently is for the cases which i checked, dynamicList, DragFloat2).

If CategoryField.ID != null the only points when appending ImGui comment to the friendlyName is when you have multiple fields on the same level with the same name.

h4cky: ooooff  i feel that explained it poorly. but basically:
if the friendlyName is "Kenki Bar Size", there is no point to append "##Kenki".
if the firendlyName is "Size", there you have a reason to append "##Kenki" to the name due to IDText() being null for for fields who are not in NestedConfig.

vond82:
https://trello.com/c/DNR4ImdD/187-only-show-when-active

vond82:
#475

vond82:

jdsmith2816 (changes requested):
Let's delay this until a beta2 patch release or beta3 and just catch them all at once.

vond82:
> Let's delay this until a beta2 patch release or beta3 and just catch them all at once.

Yeah that works, will just work on it slowly til then and hopefully we won't make it too hard to merge down the line

jdsmith2816:
@vond82 github autoclosed these PRs when I merged beta2;  please reopen these changes to beta3 branch with any relevant updates and we can get them merged in

PR #476, opened 2021-09-23
Hide overheat when not in hypercharge

zotlann:
Added an option to hide MCH overheat bar when not in hypercharge per https://trello.com/c/4gAsjKIe/179-mch-only-show-hypercharge-bar-when-hypercharge-is-used

### DIFF
--- a/DelvUI/Interface/Jobs/MachinistHud.cs
+++ b/DelvUI/Interface/Jobs/MachinistHud.cs
@@ -144,7 +144,6 @@ namespace DelvUI.Interface.Jobs
                 return;
             }
 
-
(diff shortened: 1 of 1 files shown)


Zeffuro:
[[MCH] Only show Hypercharge bar when Hypercharge is used.](https://trello.com/c/4gAsjKIe/179-mch-only-show-hypercharge-bar-when-hypercharge-is-used)

jdsmith2816 (changes requested):
These changes look fine.  Like vond's PR let's go ahead and delay this until a post Beta2 patch or beta3.  Just flagging with a change request so someone doesn't merge it.

PR #482, opened 2021-09-23
Fix DNC proc bar lag

zotlann:
Modeled the dnc proc bar after the samurai DrawSenResourceBar to reduce lag.

### DIFF
--- a/DelvUI/Interface/Jobs/DancerHud.cs
+++ b/DelvUI/Interface/Jobs/DancerHud.cs
@@ -287,7 +287,7 @@ namespace DelvUI.Interface.Jobs
             builder.SetBackgroundColor(EmptyColor.Base).Build().Draw(drawList);
         }
 
-         private void DrawProcBar(Vector2 origin)
+        private void DrawProcBar(Vector2 origin)
         {
             var statusEffects = Plugin.ClientState.LocalPlayer.StatusEffects;
             var cascadeTimer = Math.Abs(statusEffects.FirstOrDefault(o => o.EffectId == 1814).Duration);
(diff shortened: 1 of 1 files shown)


vond82:
I tried this out (tho since I'm on APIv4 I had to change some things) and there is still a delay and I can't really say if it's better or not.  I tried by having the skills that light up next to the procbar just to see the difference in when they'd light up on both builds. It's *definitely* not worse however.

One big benefit of the code in this PR compared to the current version is that it seems to once and for all fix the issues with scaling of the procbar. Thumbs up from me at least :)

DelvUI/Interface/Jobs/DancerHud.cs
```diff
@@ -289,89 +289,44 @@ private void DrawStandardBar(Vector2 origin)
...
-                showerBuilder.SetTextMode(BarTextMode.Single).SetText(BarTextPosition.CenterMiddle, BarTextType.Current);
+                if(Config.StaticProcBarsEnabled)
+                {
+                    var builder = BarBuilder.Create(cursorPos, procBarSize).
+                        AddInnerBar(procTimers[order[i]] > 0 ? 1 : 0, 1, colors[order[i]]);
+                    builder.Build().Draw(drawList);
+                }
+                else
+                {
+                    var builder = BarBuilder.Create(cursorPos, procBarSize).
```
vond82, line 323: Just insert .SetBackgroundColor(EmptyColor.Base) on the builder variables so it doesn't look out of place and I'd be happy with all of this, see my other comment

vond82 (changes requested):

jdsmith2816:
@zotlann github autoclosed these PRs when I merged beta2;  please reopen these changes to beta3 branch with any relevant updates and we can get them merged in

PR #490, opened 2021-09-25
disable compiler warnings for nullable reference types

lichie567:
This will disable huge list of compiler warnings for usage of nullable reference types. Tischel seemed okay with it, told me to create a PR and let you decide @jdsmith2816


Tischel:
We've decided to accept the nullables life.

PR #501, opened 2021-09-27
New config nesting

lichie567:
Complete re-write of the way config items nest/collapse under parent checkboxes. Allows for multiple layers of nesting collapsible config items and makes config cleaner overall.


Tischel:
Haven't gone through everything yet, but one thing I'll say already is that if we're gonna commit to a big change like this, we should also try to improve on not doing everything every frame.
Right now the "draw" of the config page nodes is ridiculously big, and the main issue is most of the things done in the draw functions are things that we could be doing only once when the nodes are created.

jdsmith2816:
@lichie567 I loaded this up locally and went through all the config panels checking things;  it seems good to me.  I looked through the code for one job config + the other files and the changes seem good to me.  

Feel free to merge it when you're happy with it and done discussing changes @Tischel

Tischel:
I'll merge this once the conflicts are fixed.
I'll look into optimizing the Draw functions eventually, there's no need for that to happen on this PR.

PR #503, opened 2021-09-27
Only show when active (meleedps)

vond82:
Started this in beta2 and was decided to hold for beta3, so here goes the first version of it that touches all meleedps-jobs.
I also decided to do some other stuff with this PR that have been discussed/have tickets while already working on the jobpacks.

- Only Show When Active-checkbox for every bar so that people can customize what they hide when not active.
- Change all instances of FirstOrDefault to find buff/debuff in the StatusList to Where.
- MNK Perfect Balance bar now makes use both of the duration and stackcount (text shown is stacks, bar itself is the duration)
- Some general cleanup, esp for MNK


vond82:

PR #504, opened 2021-09-27
WHM :: Dia bar changes

pascbeck:
- disable dia bar when not active
- pandemic refresh timer
- option to hide the timer


pascbeck:
build failed smh, pulled latest changes and re-opened.

PR #514, opened 2021-09-28
Only Show When Active for Ranged and Healers

vond82:
Continued from the previous merged PR where I did it for meleedps, this time for physical ranged and healers.

- Only Show When Active-checkbox for every bar so that people can customize what they hide when not active.
- - Where applicable. BRD Stacks bar already had similar functionality for example.
- Change all instances of FirstOrDefault to find buff/debuff in the StatusList to Where.
- DNC Esprit bar has option to unchunk as has been requested.
- DNC Proc Bar rewrite based on Jonathons beta2 PR that never was implemented.
- Implemented Slan's "The Great Restyling" as closely as possible for all touched jobs that weren't already done + for SAM.
- Some general cleanup.

### DIFF
--- a/DelvUI/Interface/Jobs/AstrologianHud.cs
+++ b/DelvUI/Interface/Jobs/AstrologianHud.cs
@@ -190,6 +190,11 @@ namespace DelvUI.Interface.Jobs
                 textSealReady = sealNumbers.ToString();
             }
 
+            if (chunkColors.Count(n => n != EmptyColor) == 0 && Config.OnlyShowDivinationBarWhenActive)
+            {
+                return;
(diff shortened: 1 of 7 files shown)


vond82:
[[MCH] Only show Hypercharge bar when Hypercharge is used.](https://trello.com/c/4gAsjKIe/179-mch-only-show-hypercharge-bar-when-hypercharge-is-used)

vond82:
[[DNC] Allow unchunking Esprit bar.](https://trello.com/c/MfXs8CXA/180-dnc-allow-unchunking-esprit-bar)

PR #515, opened 2021-09-29
Options for the new WHM bars

vond82:
- Only Show When Active option
- Track Cooldown option (not used as per default)
- Option to hide text and hiding text when duration 0
- Option for cooldown bar color


pascbeck:
👍

PR #516, opened 2021-09-29
More Primary Resource Bar options, DNC bug fixes

vond82:
- Option to hide primary resource bar when resource is full (requested)
- Option to set below threshold color for primary resource
- DNC feather gauge now properly enables when set to "only show when active" and the player gets a feather
- DNC buff bar now only shows your own buffs and not group members/dancer partners


vond82:
Converted to draft as there's an issue with the last commit that I'll need to fix in a bit

vond82:
> Converted to draft as there's an issue with the last commit that I'll need to fix in a bit

Issue fixed

PR #518, opened 2021-09-29
Only Show When Active for Tanks

vond82:
Continued from the previous merged PRs.

-Only Show When Active-checkbox for every bar so that people can customize what they hide when not active.
-- HideWhenFull-option for PLD/DRK mana.
- Implemented Slan's "The Great Restyling" where missing.
- Some general cleanup.


vond82:

PR #525, opened 2021-09-30
Minor fixes

Tischel:
* Disposing configs last to avoid race conditions
* Removing size from the party frames confg to avoid confusion

### DIFF
--- a/DelvUI/Plugin.cs
+++ b/DelvUI/Plugin.cs
@@ -255,13 +255,14 @@ namespace DelvUI
             UiBuilder.RebuildFonts();
 
             ChatHelper.Instance.Dispose();
-            ConfigurationManager.Instance.Dispose();
             FontsManager.Instance.Dispose();
             GlobalColors.Instance.Dispose();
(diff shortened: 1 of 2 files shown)


DelvUI/Plugin.cs
```diff
@@ -255,13 +255,14 @@ protected virtual void Dispose(bool disposing)
...
             ChatHelper.Instance.Dispose();
-            ConfigurationManager.Instance.Dispose();
             FontsManager.Instance.Dispose();
             GlobalColors.Instance.Dispose();
             MouseOverHelper.Instance.Dispose();
             PartyManager.Instance.Dispose();
             TexturesCache.Instance.Dispose();
             TooltipsHelper.Instance.Dispose();
+
+            ConfigurationManager.Instance.Dispose();
```
jdsmith2816, line 265:
```suggestion
            
            // This needs to remain last to avoid race conditions
            ConfigurationManager.Instance.Dispose();
```
IMO this is a good example of a place we actually want a comment

PR #536, opened 2021-10-02
Deduplicate statuses in custom effects

hintxiv:
Since "custom effects" seems to be primarily about tracking buffs, it doesn't make sense to show two icons for each raid buff that you and your target share (see image). This PR hides duplicate statuses if they were applied by the same source player.

image


Tischel (changes requested):
I think you're missing a "using" since it failed to compile.

PR #537, opened 2021-10-02
feat(partylist): added health label

Vayhyr:
preview with generic settings

### DIFF
--- a/DelvUI/Interface/Party/PartyFramesBar.cs
+++ b/DelvUI/Interface/Party/PartyFramesBar.cs
@@ -25,6 +25,7 @@ namespace DelvUI.Interface.Party
         private PartyFramesRaiseTrackerConfig _raiseTrackerConfig;
 
         private LabelHud _nameLabelHud;
+        private LabelHud _healthLabelHud;
         private LabelHud _manaLabelHud;
         private LabelHud _orderLabelHud;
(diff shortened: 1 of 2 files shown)


DelvUI/Interface/Party/PartyFramesBar.cs
```diff
@@ -58,6 +59,7 @@ PartyFramesRaiseTrackerConfig raiseTrackerConfig
             _raiseTrackerConfig = raiseTrackerConfig;
 
             _nameLabelHud = new LabelHud(id + "_nameLabel", config.NameLabelConfig);
+            _healthLabelHud = new LabelHud(id + "_nameLabel", config.HealthLabelConfig);
```
jdsmith2816, line 62: Why not call it _healthLabel ?

jdsmith2816 (changes requested):

DelvUI/Interface/Party/PartyFramesBar.cs
Vayhyr: Sorry, I copied the above and didn't change it.

Tischel:
Not sure how I feel about having 2 labels when you can just use text tags to show name and health but I guess this gives more freedom positioning wise.

jdsmith2816 (approved):

PR #547, opened 2021-10-03
fix: push fonts other than the default one

Bluefissure:
Fixs https://github.com/DelvUI/DelvUI/issues/548

### DIFF
--- a/DelvUI/Helpers/DrawHelper.cs
+++ b/DelvUI/Helpers/DrawHelper.cs
@@ -4,6 +4,7 @@ using ImGuiNET;
 using ImGuiScene;
 using Lumina.Excel;
 using System;
+using System.Collections.Generic;
 using System.Numerics;
 
 namespace DelvUI.Helpers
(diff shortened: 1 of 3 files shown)


Tischel:
PR shouldn't point to main branch.
Font work fine in beta3 api4.

Bluefissure:
> PR shouldn't point to main branch. Fonts work fine in beta3 api4.

Sorry, I thought the api3 version is still accepting PRs, I'll test it out with the latest beta in api4 to see whether the issue persists. 

> Also pushing and popping all fonts makes no sense at all.

Why bother to build them if you don't use them.

Tischel:
Sorry for the confusion, we're not pushing any updates to the api3 version since it'll will be deprecated in a week.

> Why bother to build them if you don't use them.
That's not what I meant. In you're code you're pushing every font one after the other. That's not how it should be done. You should only push the font you want to use, then pop it.

Bluefissure:
> Sorry for the confusion, we're not pushing any updates to the api3 version since it'll will be deprecated in a week.

Yeah, I noticed that, I'll check the api4 beta later and see whether this issue persists.

> > Why bother to build them if you don't use them.
> 
> That's not what I meant. In your code you're pushing every font one after the other. That's not how it should be done. You should only push the font you want to use, then pop it.
> 
> For more context, the method you're changing doesn't even exist anymore in current development.

Sure, I don't think it's a good way to simply pushing all of the fonts either. I was just confused by how the different fonts work and got surprised that none of the UI building procedures have even used that.

Anyway, I've made a compatible fix to the CN api3 plugin repo and since it's not even in the latest codebase I'll check it in the latest api4 beta.

PR #552, opened 2021-10-05
uppercase first letter in cases where its missing

vond82:
Some things like "black chocobo", "aetheryte" etc were showing in all lowercase just a quick fix to check if the string to be drawn is not empty and if so starts with a letter and if the first letter is uppercase. If not, runs totitlecase on the string

### DIFF
--- a/DelvUI/Extensions.cs
+++ b/DelvUI/Extensions.cs
@@ -4,6 +4,7 @@ using System.Numerics;
 using System.Text;
 using Dalamud.Game.Text.SeStringHandling;
 using FFXIVClientStructs.FFXIV.Client.System.String;
+using static System.Globalization.CultureInfo;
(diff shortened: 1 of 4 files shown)


Zeffuro:
Clean implementation, I like it :)

Tischel (changes requested):
This fix should not be automatically being forced on every label. What if I actually want lowercase?.
If this is a bug with action names in cast bar, do this when setting the text of the label config in the castbar hud.

vond82:
> This fix should not be automatically being forced on every label. What if I actually want lowercase?. If this is a bug with action names in the cast bar, do this when setting the text of the label config in the castbar hud.

It's also for names on unitframes but yeah I'll check for another solution after wowraid

PR #559, opened 2021-10-07
fix: WHM Asylum not working sub lvl78

vond82:
Also some general cleanup


Tischel:
Gettings this in before merging the new bar stuff

PR #607, opened 2021-10-10
Changelog

Tischel:
* Changelog window shows up automatically the first time a new version is loaded
* Cleaned up config window layout and added a changelog button
* Fixed crash when opening browser

image
image

### DIFF
--- a/DelvUI/changelog.md
+++ b/DelvUI/changelog.md
@@ -1,41 +1,44 @@
 # 0.3.0.0
-- Added changelog :)
-- This version brings A LOT of changes. Chances are your settings will be invalid / broken. We recommend starting from a clean state.
-- Added more textags.
-- Fixed several issues in the HUD unlocked mode.
-- Added mana bars for Target, Target of Target and Focus Target.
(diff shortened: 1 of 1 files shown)


Tischel:
> I really like it! Only thing I'd consider would be having the "big stuff" earlier in the changelog such as partyframes and profiles. Maybe a "highlights" section?

Ah yea, that's a good idea.
I kinda did it in chronological order but prob better to re-organize it a bit.

Tischel:
Re-organized the changelog a bit:
https://github.com/DelvUI/DelvUI/blob/f79a8fb18a1729da2d7a42c6472dea919394b438/DelvUI/changelog.md

Kanubiz:
Hmm i would probably change the style of the hide and unlock hud buttons if we're leaving a clean mask on the changelog button

Tischel:
> > Would also add a "See what's new!" Headline to the changelog
> 
> And format it a little better, can use the ascii icon i used for the collapsibles instead of the -

It's markdown

PR #692, opened 2021-10-15
Show in duty (when Hide DelvUI outside of combat is selected)

vond82:
option to always show delvui frames in duties when "hide delvui outside of combat" is selected


vond82:
Changing this to draft cause I realized I was too tired while doing this and forgot what the feature actually SHOULD do, will sort it later today when i have a few spare minutes

PR #693, opened 2021-10-15
Added Dead Backdrop option

vond82:
Added option to change the background color of dead players on unitframes and party frames

About the ugly way of using HP is 0, after looking thru the dalamud API documentation I checked the PlayerTracker plugin as it has the option to disable nameplates for dead players. It used a IsDead() method that came from a Dalamud extension library (DrunkenToad) and all that method did was check if currenthp == 0 so I'd say this method is "good enough" for now


Tischel:
Be aware that this will also happen on cross world parties or for players in other areas where you dont have the actors in the actor table.
HP will be 0 then as well.

Kanubiz:
> > Be aware that this will also happen on cross world parties or for players in other areas where you dont have the actors in the actor table. HP will be 0 then as well.
> 
> Personally I'd definitely find that fine at least

I agree with vond

Issue #678, opened 2021-10-14
[Suggestion]: Bring back chunked bar number text


Necroskillz:
### What's your suggestion?

In beta 3, the possibility to show labels on bar chunks has been intentionally removed. I'd like to propose to re-add this feature. Maybe it's just me, but I feel more comfortable when I see the numbers.

On MCH I like having chunked bar for heat, but I'd like to know how many GCDs I'm away from being able to hypercharge or how close am I to overcapping, and having a number is more readable then estimating from a bar.

In case of DNC esprit bar, you use saber dance on 80 (outside of tech step), so it's even more important to see the number (esp if you get to some weird number with improv). Here maybe it would make sense to remove chunking entirely and change it to a normal bar (like MCH battery).

Just putting this here after some comments in discord, so ppl can add their use cases and see if more ppl miss this feature as well.

PR #694, opened 2021-10-15
Re-enable labels on chunked bars

Necroskillz:
Impl of #678. The functionality to display labels was already there, but unused. This reactivates it.


vond82:
Please direct PRs to the develop-branch. Also this was disabled on purpose, but make the PR and we'll consider it

PR #695, opened 2021-10-15
Re-enable labels on chunked bars

Necroskillz:
Impl of #678. The functionality to display labels was already there, but unused. This reactivates it.


vond82:
I'm not against the idea of adding the option as long as we make sure it's not used by default anywhere (except possibly Mudra)

Tischel:
@Necroskillz can you show how it looks, specifically the values for all the labels

PR #738, opened 2021-10-17
Rewrite ConfigPageNode generation

lichie567:
- re-wrote the config building logic from the ground up. mainly did this to allow NestedConfig objects to be drawn between non-nested config fields (before, NestedConfigs were always drawn at the end of a config page, regardless of order specified)
- added a config attribute for RadioSelector

Overall the config code is 100x less spaghetti-like now. I'm planning to do some more config stuff to make use of this functionality soon.


Tischel:
@lichie567 made a pr to this branch with the config merging stuff https://github.com/lichie567/DelvUI/pull/1

PR #757, opened 2021-10-20
Role colors and icons for unit frames

Tischel:
* Color health by value extracted to its own config
* Added appropriate converters to port them
* Extracted role colors from party frames to global colors
* Added role color setting to unit frames
* Added role/job icon to unit frames
* Added setting in unit frames to hide any health labels if the health of the unit is invalid
* Some minor unit frames related fixes


Tischel:
@vond82 should work now

PR #758, opened 2021-10-20
Utilize partial fill for chunked bars with disabled chunking

Necroskillz:
`PartialFillColor` is currently not used at all when `UseChunks` is `false`. This uses it in the following way: partial fill means that the first chunk (if the bar was chunked) wasn't filled.

I included a variable for which chunk needs to be filled in order to not be considered partialy filled, because maybe for some job it might make sense to have it differently (e.g. partial fill for 2 out of 3 chunks etc.). Can remove this if it's not needed and hardcode to 1 as well.

Example (DNC esprit)
image
image

### DIFF
--- a/DelvUI/Interface/Bars/BarUtilities.cs
+++ b/DelvUI/Interface/Bars/BarUtilities.cs
@@ -238,7 +238,8 @@ namespace DelvUI.Interface.Bars
             float min = 0f,
             GameObject? actor = null,
             BarGlowConfig? glowConfig = null,
-            PluginConfigColor? fillColor = null)
+            PluginConfigColor? fillColor = null,
+            int thresholdChunk = 1)
         {
             var color = fillColor ?? config.FillColor;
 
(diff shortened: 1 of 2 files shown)


Necroskillz:
This is just an adapter for the threshold. Its always below, because its partial fill.

I thought that it would be possible to use Partial Fill Color for chunks, and the standard Threshold config for unchunked, but afaik you can't switch between the two based on a checkbox in the config (only hide if checkbox is unchecked with `collapseWith`).

Tischel:
@Necroskillz There should be a setting to disable this (no chunks + no partial fill color should be doable since its the current behavior and we don't wanna remove that option)

Necroskillz:
> @Necroskillz There should be a setting to disable this (no chunks + no partial fill color should be doable since its the current behavior and we don't wanna remove that option)

There is, just uncheck `Use Partial Fill Color`

Tischel:
oh, im dumb

PR #809, opened 2021-10-30
Adding start angle stting for gcd indicator

Tischel:


DelvUI/Interface/GeneralElements/GCDIndicatorConfig.cs
```diff
@@ -54,6 +54,10 @@ public class GCDIndicatorConfig : AnchorablePluginConfigObject
         [Order(40, collapseWith = nameof(CircularMode))]
         public int CircleThickness = 10;
 
+        [DragInt("Start Angle", min = 0, max = 359)]
```
vond82, line 57: Maybe add a tooltip to indicate its a 360 degree value so people don't try to start from "6 o'clock" etc, but that might be overdoing it. Otherwise all good!

vond82 (approved):

DelvUI/Interface/GeneralElements/GCDIndicatorConfig.cs
Tischel: eeee an angle is an angle xd

vond82: Yeah I just dont want that as a question popping up but let's trust that ppl get it from just playing with the value :D

PR #812, opened 2021-10-31
try on screen and duty leave menu + sneak 1 more fellowship menu

Caliel666:
first pr pog


Caliel666:
added the examine window as well

PR #829, opened 2021-11-04
Add wago button

lichie567:


lichie567:
I left the button styled the same as the changelog button. If someone else wants to re-style it to use wago colors go ahead

Issue #828, opened 2021-11-03
[Suggestion]: pet buffs/debuffs shown as "only my effects"


Caliel666:
due to popular demand

PR #830, opened 2021-11-04
Implement pet effects as own #828

Zeffuro:
Implements option to set pet effects as your own effects. Implements #828 

https://puu.sh/ImXLw/122d9af816.mp4

### DIFF
--- a/DelvUI/Interface/StatusEffects/StatusEffectsListHud.cs
+++ b/DelvUI/Interface/StatusEffects/StatusEffectsListHud.cs
@@ -182,7 +182,14 @@ namespace DelvUI.Interface.StatusEffects
                 }
 
                 // only mine
-                if (Config.ShowOnlyMine && player?.ObjectId != status->SourceID)
+                var mine = player?.ObjectId == status->SourceID;
+
+                if (Config.IncludePetAsOwn)
+                {
+                    mine = player?.ObjectId == status->SourceID || IsStatusFromPlayerPet(*status);
+                }
+
+                if (Config.ShowOnlyMine && !mine)
                 {
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


DelvUI/Interface/StatusEffects/StatusEffectsListHud.cs
```diff
@@ -199,6 +206,28 @@ protected unsafe List<StatusEffectData> StatusEffectDataList(GameObject? actor)
...
 
+        protected bool IsStatusFromPlayerPet(StatusStruct status)
+        {
+            var player = Plugin.ClientState.LocalPlayer;
+            if (player == null)
+            {
+                return false;
+            }
+            
+            var character = Plugin.ObjectTable.SearchById(status.SourceID);
```
Tischel, line 217: Doing this every frame for every status for every list is pretty costly.
I think you can directly find the pet's address/id by using the BuddyList class from Dalamud, but haven't tested it.

Zeffuro: Well the only other options I thought off is including it in the StatusEffect struct but that required a whole more bunch of changes to if everywhere so wasn't sure what option would be best.

Tischel: I think the way you did it is fine if we can simplify the look up for the pet's actor.
I'm pretty sure you can get it directly with the class I mentioned above, but again, haven't really tested it.

Zeffuro: Actually that was exactly what I needed, thanks for bringing that class up, works for me :)

Tischel (approved):

PR #836, opened 2021-11-05
Fixing sorting crash on status effects lists

Tischel:


Zeffuro:
Tested basic functionality and everything still seems to work like expected.

Zeffuro (approved):

PR #854, opened 2021-11-20
Fixing nesting issues with the new collapsing headers

Tischel:


vond82 (changes requested):
I added a PR that fixes a small annoyance with the SAM config, other than that all good from me

PR #857, opened 2021-12-06
[Misc] Fix Has Weapon Drawn Check

d0tiKs:
Replace signature and check for the method HasWeaponDrawn, using ClientState.LocalPlayer and its StatusFlags to return the result

Edit : The branch weirdness is because I had to remove the auto formating of my IDE


Tischel:
Was there a bug in the code or is this just for clarity?

d0tiKs:
Both, I compiled the api5 version for the current release of the game and the feature didn't work anymore. 

Also, recently I wanted this same exact feature for another plugin, and after some advice from the Dalamud/XIVLaunched devs on discord, and noticing that this feature was still working with EW, I implemented it to delvUI using the solution I used on this other plugin.

PR #859, opened 2021-12-10
6.0 Summoner Rewrite.

TheResinger:
https://www.youtube.com/watch?v=A6WMkAduyLE

### DIFF
--- a/DelvUI/Interface/Jobs/SummonerHud.cs
+++ b/DelvUI/Interface/Jobs/SummonerHud.cs
@@ -1,6 +1,5 @@
 using Dalamud.Game.ClientState.JobGauge.Types;
 using Dalamud.Game.ClientState.Objects.SubKinds;
-using Dalamud.Logging;
 using DelvUI.Config;
 using DelvUI.Config.Attributes;
 using DelvUI.Enums;
@@ -100,10 +99,6 @@ namespace DelvUI.Interface.Jobs
             }
 
             HandleAttunementStacks(pos, player);
-            // if (Config.RuinBar.Enabled)
-            // {
-            //     DrawRuinBar(pos, player);
-            // }
 
         }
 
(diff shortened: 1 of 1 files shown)


TheResinger:
[SMN HUD Update](https://trello.com/c/0iD8kU1A/223-smn-hud-update)

DelvUI/Interface/Jobs/SummonerHud.cs
```diff
@@ -83,171 +78,173 @@ public override void DrawJobHud(Vector2 origin, PlayerCharacter player)
...
+                    }
+                    break;
+                default:
+                    if (Config.StacksBar.Enabled && !Config.StacksBar.HideWhenInactive)
+                    {
+                        DrawStacksBar(origin, player, 0, 1, Config.StacksBar.FillColor);
+                    }
+                    break;
             }
+        }
```
Tischel, line 159: 1) Any reason why you didn't check for `Config.StacksBar.Enabled` before even calling this method?
2) The `currentAttunement` variable is really weird. Making it a string and then doing a switch over it is weird. IMO just do some if-elses i.e.:
```
if (gauge.IsIfritAttuned && Config.StacksBar.ShowIfritStacks) 
{
    // ...
}
else if (gauge.IsTitanAttuned && Config.StacksBar.ShowTitanStacks)
{
    // ...
}
// etc...

Tischel (changes requested):

DelvUI/Interface/Jobs/SummonerHud.cs
TheResinger: Brain lag is one hell of a drug. 

1. Meant to implement the check but skimped over it on my final check, added the check
2. I agree with this. rewrote this section has a chain of else ifs.

PR #860, opened 2021-12-12
DNC Proc Bar Fix

TheResinger:
Buff IDs for the procs got updated.


TheResinger:
[DNC HUD Update](https://trello.com/c/pbn9Rznp/221-dnc-hud-update)

Issue #851, opened 2021-11-18
[Suggestion]: Different color for max HP when using color based on health values


boratintheuk:
### What's your suggestion?

In ElvUI when setting unit frames to use color based on health value it goes from green to red, but at full health it goes gray.

Staring at green unit frames can be straining after awhile and a nice neutral color like gray would make them easier to look at.

While you could set full health color to gray or another neutral color with the current DelvUI setting, you would not a get a proper color scale.

Maybe just add a third color option? Low/High/Full instead of just Low/Full

Tischel:
Implemented in 0.6.0.0

PR #863, opened 2021-12-13
Allow 3rd color option when health is above max health threshold

lemonlust:
https://github.com/DelvUI/DelvUI/issues/851#issue-1056987296

image


lemonlust:
close to reevaluate method

PR #878, opened 2021-12-15
fixed bard dot durations.

TheResinger:
Dot durations were updated to 45s each for both versions of each dot.


Tischel:
I completely missed this PR and did myself in another one.
Im stupid, sry.

PR #890, opened 2021-12-16
Warrior Hud: Added Primal Rend Glow

sebkoenig:
Added optional glow on Inner Release bar when Primal Rend is active.


Tischel (changes requested):
Please add an entry in the changelog under "Features" for version 0.6.0.2

sebkoenig:
Updated the changelog.

PR #905, opened 2021-12-19
[DONT MERGE] bar strata

lichie567:


Tischel:
Can't try it right now cuz queue but just by looking at the code it doesn't seem like this would work for all cases.

There's a lot of hud elements that draw bars and stuff on top of it. If im understanding this correctly, the bar will be drawn last now.
To give an example, in the enemy list we draw the enmity icons right after drawing the bars. How does this work now?

lichie567:
yeah this was just cherry picked from my old branch. i don't think we had icons drawn over bars back then. i'll have to come up with something

Tischel:
Closing this since I'm working on the complete feature, will open a PR soon (tm)

PR #907, opened 2021-12-19
Added a Minor Arcana Bar for Astrologian.

vond82:


Tischel (changes requested):
Missing "Hide When Inactive" functionality

vond82:
> Missing "Hide When Inactive" functionality

Nah, line 327 :)

PR #911, opened 2021-12-20
fix possible invalidcastexception in partycooldownsmanager

vond82:


DelvUI/Interface/PartyCooldowns/PartyCooldownsManager.cs
```diff
@@ -167,8 +167,10 @@ private void OnMembersChanged(PartyManager sender)
             else if (!_config.ShowOnlyInDuties || Plugin.Condition[ConditionFlag.BoundByDuty])
             {
                 // add new members
-                foreach (PartyFramesMember member in sender.GroupMembers)
+                foreach (IPartyFramesMember partyFramesMember in sender.GroupMembers)
                 {
+                    PartyFramesMember member = (PartyFramesMember)partyFramesMember;
+
```
Tischel, line 173: You shouldn't need to do this, the other change is correct though

Tischel (changes requested):

DelvUI/Interface/PartyCooldowns/PartyCooldownsManager.cs
vond82: The reason for that was to make as few changes as possible, as otherwise there's several methods that needs changing (CooldownsForMember etc etc)

Tischel:
Closing since this change will be in the other PR

PR #915, opened 2021-12-20
 Fixing actor reference on labels for all dot and proc bars, and BLM's enochian bar

Tischel:


Tischel:
Random blue mage editor config clean up /shrug

vond82 (approved):

PR #925, opened 2021-12-23
fix missing health with transparent bars

lichie567:


vond82:
works, tested both with and without tank invuln active

PR #927, opened 2021-12-23
Adding party cooldowns tooltips

Tischel:
Can't show the actual spell descriptions because the data we get is not very user friendly / readable.
Couldn't figure out how to deal with so for now it just shows name, duration and recast time.

Also adding rescue, swiftcast and tank invulns as tracked cooldowns

### DIFF
--- a/DelvUI/Interface/PartyCooldowns/PartyCooldownsConfig.cs
+++ b/DelvUI/Interface/PartyCooldowns/PartyCooldownsConfig.cs
@@ -1,4 +1,5 @@
-﻿using DelvUI.Config;
+﻿using Dalamud.Utility;
+using DelvUI.Config;
 using DelvUI.Config.Attributes;
 using DelvUI.Enums;
 using DelvUI.Helpers;
@@ -43,6 +44,10 @@ namespace DelvUI.Interface.PartyCooldowns
         [DragInt2("Padding", min = -1000, max = 1000)]
         [Order(15)]
         public Vector2 Padding = new Vector2(0, -1);
+
+        [Checkbox("Tooltips", spacing = true)]
+        [Order(16)]
+        public bool ShowTooltips = true;
 
         [Checkbox("Show Only in Duties", spacing = true, isMonitored = true)]
         [Order(20)]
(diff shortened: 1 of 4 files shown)


DelvUI/Interface/PartyCooldowns/PartyCooldownsConfig.cs
```diff
@@ -368,35 +367,57 @@ private void DrawFilter(string name, JobRoles role)
...
+
+
+            // SINGLE-TARGET EFFECTS (disabled by default)
+
+            [30] = NewData(30, JobIDs.PLD, 50, 420, 10, 100, 4, false), // hallowed ground
+            [43] = NewData(43, JobIDs.WAR, 42, 240, 10, 100, 4, false), // holmgang
+            [3638] = NewData(3638, JobIDs.DRK, 50, 300, 10, 100, 4, false), // living dead
+            [16152] = NewData(16152, JobIDs.GNB, 50, 360, 10, 100, 4, false), // superbolide
+            [7571] = NewData(7571, JobRoles.Healer, 48, 120, 0, 80, 4, false), // rescue
+            [7561] = NewData(7561, new List<JobRoles>() { JobRoles.Healer, JobRoles.DPSCaster }, 18, 60, 0, 80, 4, false), // swiftcast
```
vond82, line 379: swiftcast actually has a duration of 10s

vond82 (changes requested):

DelvUI/Interface/PartyCooldowns/PartyCooldownsConfig.cs
Tischel: I'm not tracking the actual buff and Swiftcast is almost always used instantly. I don't really think there's much value in tracking the actual duration for this one. I could just set it to 10 duration but it wont be have accurately unless I add a layer to track the actual buff duration.

vond82: Alright

vond82 (approved):

PR #937, opened 2021-12-30
Strata levels

Tischel:
* StrataLevels is a property of MovablePluginConfigObject
* HUDManager manages the first "layer" of strata levels by having an ordered array of the base UI elements sorted by strata level
* Each HUDElement has a second "layer" of strata levels by having ordered arrays of its sub elements sorted organized by strata level
* Party Frames were giving me a lot of issues. I removed the "party frames are unlocked, draggable and resizable when config window is opened" system. It caused many issues even before this implementation. The Position, Rows and Columns settings can be used instead to define the layout of the Party Frames.

Note: Party Frames still need to be ported to BarHud. Will probably do that in another PR for that once this gets merged.

### DIFF
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -228,7 +228,7 @@ namespace DelvUI.Config.Attributes
     {
         public int min;
         public int max;
-        public int velocity;
+        public float velocity;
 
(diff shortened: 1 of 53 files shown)


vond82:
Seems stable and working enough so merging into develop so we can build from this, more fixes can be made before release still anyway if need be

vond82 (approved):

PR #938, opened 2021-12-30
Castbar stuff

vond82:
- Added option to change fill direction of the castbar. Right is still the default.
-- If FillDirection is Left then Slidecast position is moved to account for this. If Ability Icon is also shown then this is accounted for as well.
- Added option to use ReverseFill in place of normal background color much like the Missing Health color does for unitframes and enemy frames.

### DIFF
--- a/DelvUI/Interface/GeneralElements/CastbarHud.cs
+++ b/DelvUI/Interface/GeneralElements/CastbarHud.cs
@@ -8,7 +8,6 @@ using FFXIVClientStructs.FFXIV.Client.Game;
 using ImGuiNET;
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Numerics;
 
 namespace DelvUI.Interface.GeneralElements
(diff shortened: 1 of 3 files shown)


DelvUI/Interface/GeneralElements/CastbarHud.cs
```diff
@@ -55,24 +51,35 @@ public override void DrawChildren(Vector2 origin)
...
 
-            UpdateCurrentCast(out var currentCastTime, out var totalCastTime);
+            UpdateCurrentCast(out float currentCastTime, out float totalCastTime);
             if (totalCastTime == 0)
             {
                 return;
             }
 
-            bool validIcon = LastUsedCast != null && LastUsedCast.IconTexture != null;
+            bool validIcon = LastUsedCast is { IconTexture: { } };
```
Tischel, line 60: I hate this with passion. Prob because I'm new to the language, but I would have literally no idea what this meant if I didn't have the original line there.

Not sure if there's a name for this kind of syntax and not sure if you're doing it manually or it's some automatic style settings you have, but I REALLY don't want it lol

Tischel (approved):

Tischel:
Don't forget about changelog!

DelvUI/Interface/GeneralElements/CastbarHud.cs
vond82: I don't mind changing it back, I just tend to randomly change stuff like that when thinking about something (bad habit prolly haha). It's just (property) pattern matching anyway, but yeah, != null or is not null has way better readability so ill change it back and add changelog in a bit

PR #952, opened 2022-01-03
font scopes

lichie567:
idk i was bored


Tischel:
Are you 100% sure the dispose gets called when it gets out of scope? Otherwise it could cause unbalanced push/pops and that destroys ImGui.

lichie567:
yeah thats exactly how the using keyword works.
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement

PR #955, opened 2022-01-04
Job role background color for party/unitframes

vond82:
- Added option to use Job and Role Color as Background Color in Party Frames.
- Added option to use Job and Role Color as Missing Health Color in Party Frames.
- Added option to use Role Color as Background Color in Unit Frames.
- Added option to use Job and Role Color as Missing Health Color in Unit Frames.
- Fixed Party Frames Tank Invulnerability Background Color when using Missing Health Color.


vond82:
Converting to draft as I'm gonna add some more stuff to it, just need to verify the blindcoding works once I'm thru queue

vond82:
Finally got ingame and can confirm this works now :P

PR #969, opened 2022-01-10
Extra Custom Effects

vond82:
By popular request, disabled by default. Not pre-populated like the first one etc.


Tischel:
I don't like this

PR #984, opened 2022-01-15
separate automaton/rook from battery gauge

vond82:
separated automaton/rook duration from battery gauge to its own bar
since i dont have a mch over 30 it would be nice if @Zeffuro or someone else that can at least summon rook automaton could test this


Tischel (approved):

vond82:
Closing this as i kinda deleted the local branch before it was merged haha. Will push a new one with a fix and merge it.

PR #995, opened 2022-01-26
Label number handling (truncate, floor, ceil, round)

Zeffuro:
There's probably a better way to name/describe this option, but my brain couldn't find it at the moment.

I mostly added this so people can set it to ceil to make it match what the game is doing for cooldowns for the trance bar for example.

https://user-images.githubusercontent.com/4972345/151149686-480b7b85-1343-431a-ae51-93694653af70.mp4


Tischel:
I'd prob name it "Rounding Mode" or something like that.

Issue #1013, opened 2022-02-08
[Suggestion]: Make decimal health percentage digit count consistent


BerkeBat:
### What's your suggestion?

Currently when using the decimal health percentage tag for something like unitframes, whenever the percentage ends up on a whole number, there is no decimal .0 shown. For example it will go from 75.1% to 75% then to 74.9%. Obviously the value is correct but if there is other text in the label, the final digit disappearing and reappearing causes the rest of the text to move left and right all the time. Not a huge issue but a small QOL change would be to change it so that it goes from 75.1% to 75.0% then to 74.9%.

I think 100% should still stay without the .0 and I'm not sure about 0%, that's probably fine too since it will only reach that once and won't cause the text movement during combat

vond82:
I don’t agree that this is an improvement, but could possibly add it as a separate texttag

PR #1014, opened 2022-02-08
Added new tag for consistent digit health percentage (#1013)

BerkeBat:
This is in relation to #1013 
Would be great to get some feedback on:
- If the new tag name is clear `[health:percent-decimal-consistent]`.
- If the place I put the helper method `ConsistentDigitPercentage` is okay (at the bottom of TextTagsHelper.cs)
- If I should also add the tag to the mana dictionary.


Tischel:
I don't see a problem adding this, but I feel like the name of the tag is not very intuitive.
At the same time, I can't come up with a better name for it lol

vond82:
> [health:percent-decimal-uniform]

I vote for this one. @Tischel you ok with this?

vond82:
> last thing should I add a similar tag in for mana for consistency?

sure!

PR #1035, opened 2022-04-15
 Fixing shields

Tischel:


Zeffuro:
Works fine, tested with multiple classes.

PR #1059, opened 2022-05-06
Added sub-option to Hide DelvUI outside of combat to show when crafting

chalkos:


chalkos:
I also noticed `isHidden` was doing a redundant check, so I improved that.

PR #1063, opened 2022-05-10
Gathering nodes hp

Tischel:
* Only changes the actual health bar
* No text tags support (its a lot of effort to get that working)


Zeffuro:
Tested a bunch of times and it works fine for me :)

PR #1067, opened 2022-05-17
Option to show DelvUI when gathering. Alongside matching options for ActionBars

SeanLatimer:
Added a gathering companion to the new show UI when crafting setting. Also added matching options for both on the ActionBars side.

### DIFF
--- a/DelvUI/Interface/HudHelper.cs
+++ b/DelvUI/Interface/HudHelper.cs
@@ -193,6 +193,10 @@ namespace DelvUI.Interface
                 isHidden = isHidden && Config.ShowDelvUIFramesWhenCrafting
                     ? Config.HideOutsideOfCombat && !IsInCombat() && !IsCrafting()
                     : isHidden;
+                
+                isHidden = isHidden && Config.ShowDelvUIFramesWhenGathering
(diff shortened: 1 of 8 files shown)


DelvUI/Interface/HudHelper.cs
```diff
@@ -271,9 +275,15 @@ private void UpdateCombatActionBars(bool forceUpdate = false)
...
-                        : IsInCombat();
+                    : currentCombatState;
+                
+                currentCombatState = !currentCombatState && Config.ShowCombatActionBarsWhenCrafting
+                    ? IsCrafting() || IsInCombat()
+                    : currentCombatState;
+                
+                currentCombatState = !currentCombatState && Config.ShowCombatActionBarsWhenGathering
+                    ? IsGathering() || IsInCombat()
+                    : currentCombatState;
```
Tischel, line 286: This doesn't seem right.
ShowCombatActionBarsInDuty and IsInDuty are not being used anymore unless I missed something.

Tischel (changes requested):

Tischel:
The whole visibility settings are being reworked for v1.2.0.0
I added this functionality there.

PR #1081, opened 2022-08-03
Implement Bar Shadow Options

Zeffuro:
The thing nobody asked for :3

ffxiv_dx11_2022-08-03_17-33-45

### DIFF
--- a/DelvUI/Interface/Bars/BarHud.cs
+++ b/DelvUI/Interface/Bars/BarHud.cs
@@ -13,6 +13,8 @@ namespace DelvUI.Interface.Bars
     public class BarHud
     {
         private string ID { get; set; }
+        
+        private BarConfig? Config { get; set;}
 
         private Rect BackgroundRect { get; set; } = new Rect();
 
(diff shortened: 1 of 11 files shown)


DelvUI/Interface/Bars/BarHud.cs
```diff
@@ -13,6 +13,8 @@ namespace DelvUI.Interface.Bars
     public class BarHud
     {
         private string ID { get; set; }
+        
+        private BarConfig? Config { get; set;}
```
Tischel, line 17: BarHud was coded in such a way so it doesn't depend on a config object, that's why it has a constructor that receives the raw values of all parameters, and a convenience constructor that receives a config and extracts the values.
The idea of this is that you don't necessarily need a config to draw a bar.

I think the ideal way to implement this would be to create a ShadowConfig object and add it as nested object of the HUD elements you want shadows on. This would also mean modifying LabelConfig so it now has a ShadowConfig instead of the shadow related variables directly in it.
That way you can make the BarHud constructor take a ShadowConfig (instead of the whole BarConfig) since that's really all you need.

Zeffuro:
Made some more changes

PR #1088, opened 2022-08-27
added 'formatted' tag for current and max HP and MP

BerkeBat:
Added a `xxx:current-formatted` and `xxx:max-formatted` tag for both HP and MP that adds a thousands separator to increase readability of large numbers. e.g. `136619` -> `136,619`.

I got 'formatted' from the discord suggestions channel but I'm happy to change it if it isn't clear.


Zeffuro:
Looks good but was wondering if you could possibly look into making this implementation for more completeness 
https://i.imgur.com/zZgHq2n.png

BerkeBat:
I don't think I have the C# skills required for something like that 😅. If I'm understanding what you mean, that is.

PR #1152, opened 2023-08-18
Add Devilment and Standard Step

Zeffuro:
Adds Devilment and Standard Step, Standard Step is disabled by default.

Ideally I'd like to add AST Cards and BRD Songs as well but dunno how since we completely disregard actual buffs on people.

Technically Devilment is only active on one person but since we already have Dragon Sight we should be consistent and have it too but for future cleanup/correctness we should check if the effect is on the player (or make it an option to only show stuff active on you). 

When we have something that also checks the buffs in place I could move and add Songs and Cards


Tischel:
I don't think standard step belongs here as its something that is (or should) be permanently up.

I'm closing this PR but I'll add Devilment myself.

PR #1183, opened 2024-06-04
Attempted to fix the colors for Honorific.

FoxMaccloud:
I have tried to implement colors for the Honorific plugin.
I have not contributed much to open source before, but this was a feature I wanted which I found missing in Delvui.
To accomplish this I have added some fields to classes, instead of handling some of the IPC data, it will now handle all of it. And I also needed to make a bit of changes to the draw functions.  I just made a new function called DrawTitle which takes a glow argument instead of making changes to DrawLabel, but they are mostly the same.
I'm not very much of a C# developer, so it was a bit of fun trying something new. I hope the code isn't too bad, but if you see something you want me to change, I'm happy to follow up on it.

### DIFF
--- a/DelvUI/Helpers/HonorificHelper.cs
+++ b/DelvUI/Helpers/HonorificHelper.cs
@@ -6,6 +6,7 @@ using Newtonsoft.Json;
 using Lumina.Data.Parsing.Uld;
 using System;
 using System.Linq;
+using System.Numerics;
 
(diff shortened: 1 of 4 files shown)


Zeffuro:
This PR has been pretty stale now and I know from discussing this before that we did not like the implementation back then.
We still might want the functionality but done in a slightly better way. 

Closing this for now.

PR #1197, opened 2024-07-07
Remove drawing from PCT gauge

trevorm4:
Currently, the middle bar in the PCT canvas gauge uses the `CreatureFlags` to determine which color to draw in the middle. The CreatureFlags struct simply populates which drawing have been used already, which ends up in practice making the middle bar always equal to the MoogleColor because the moogle if statement is first, or an empty bar if the current painting is moogle.

With moogle painted
image

With claw painted
image

With fang painted
image

You can see that despite our last painting being a Fang muse, the middle bar is always the moogle color which is not very helpful information

In practice, this gauge is essentially useless because the previous Creature can always be inferred from the current canvas Gauge (1st bar).

In the base FF ui, the hud looks like
image

The top part of the gauge is what the middle bar was attempting to mimic, but there isn't a way to convey this information in a single bar because multiple flags can be active for the different drawings that have been consumed

Before
image

After
image

### DIFF
--- a/DelvUI/Interface/Jobs/PictomancerHud.cs
+++ b/DelvUI/Interface/Jobs/PictomancerHud.cs
@@ -199,6 +199,26 @@ namespace DelvUI.Interface.Jobs
                 null
             );
 
+            // part drawing
+            PluginConfigColor? drawingColor = null;
+            if (gauge.CreatureFlags.HasFlag(CreatureFlags.Pom))
+            {
+                drawingColor = config.PomColor;
+            }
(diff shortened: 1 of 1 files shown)


Tischel:
Having the middle part can be useful for some people to know if they're getting moogle or madeen next.
I agree it's not very important but I don't want to remove it.
I'll fix the logic so the color represents the last creature part drawn and I'll add a flag to have this part of the bar be optional.

Thanks for your feedback!

PR #1199, opened 2024-07-08
feat(viper): adds `invert` option to vipersight guage

cultbaus:
What: Adds an invert checkbox to the `Vipersight` gauge

Why: As a user, my `Dread Fangs` keybind is `Q`, my `Steel Fangs` keybind is `E`, with the default Vipersight gauge orientation, when the left-most side glows to indicating the finisher to execute, I must press the right-most key. With this checkbox, I can better align my physical keypresses with the Vipersight gauge information.


Zeffuro:
Looks good to me, I'll probably play with inverted myself because Steel Fangs on my 2 as the more "default" button during downtime makes more sense to me :)

PR #1203, opened 2024-07-08
Added more TextTags

PameiaB:
Added Text Tags for solely the first and last initial individually and rewrote initials to utilize them to avoid repeat code


PameiaB:
Accidentally closed

Tischel:
This should be coded into the text tags system instead of creating an individual text tag for the last letter.

https://github.com/DelvUI/DelvUI/wiki/Text-Formatting

Example:
`[name:first.1]` gives the first letter (already possible)
`[name:first.-1]` would give the last letter (new feature)

PameiaB:
apologies, I was unaware there was a wikipage and didn't not know about this feature

Issue #1207, opened 2024-07-09
[Suggestion]: Allow Vipersight gauge color to change based on buff


Blitzkind:
### What's your suggestion?

Currently the Vipersight gauge has two sections denoting the step of the combo you're in, and fills the gauge based on that colors you signify in settings. These colors (Default Blue and Red) do not have any bearing on what positional you need to fulfill.

I suggest that an option be added where the color of the final section or the glow that highlights the correct button be dependent on what buff you have. For example, if I have the buff that increases potency for rear positional finishers it'll fill the last section with one color, and if I have the buff for flank positional finisher it would be the other.

Currently the only way to know his at a glance is to look at the hotbar or buff bar. Worse yet, if you are Red-Green colorblind, it's even harder to know. Giving the ability for users to pick a color to signify this would be a great accessibility tool.

Edit: Screenshots to explain what I mean and what I want, hotbars modified to give info.

New Combo, no input
![image](https://github.com/DelvUI/DelvUI/assets/5423920/6171f42b-ee7b-43cd-ae7c-1b84295aa7ab)

Press Steel Fangs
![image](https://github.com/DelvUI/DelvUI/assets/5423920/93041ed1-47d3-483e-8843-f237ecdf0621)

Pressed Hunter's Sting
![image](https://github.com/DelvUI/DelvUI/assets/5423920/e41ca2d4-12ff-46f8-892d-0a33519e1dd9)

What I want here: Allow me to change the color of either the final section of the gauge or the glow to reflect that I need to do a flank positional as both options here are flank positional attacks. 

Likewise, I would like this to reflect the same for rear as I continue the combo as shown below

New combo, Rear buff
![image](https://github.com/DelvUI/DelvUI/assets/5423920/09957029-e8af-4b7a-b180-09d0e4e9aec9)

Second Step, Rear buff
![image](https://github.com/DelvUI/DelvUI/assets/5423920/fcfd5e67-0cdc-432b-bc75-0ba82ff4df89)

Press Swiftskin's Sting
![image](https://github.com/DelvUI/DelvUI/assets/5423920/1b4259fe-9356-4d76-bf61-57411dffa242)

What I want here: Allow me to change the final fill in color to be distinct from the first combo, such that I can know at a glance what positional I need to fulfill as both attacks here are rear positionals.

cultbaus:
The vipersight gauge has a glow indicating which combo action will consume your currently active buff, if both sides are glowing that means that both finishers are valid.

[...]

PR #1208, opened 2024-07-10
Added Vipersight feature: #1207 Allow Vipersight gauge color to change based on buff

Blitzkind:
Adds an option to specify what color the final section of the Vipersight bar will be based on available combo finshers:

image

image

image

### DIFF
--- a/DelvUI/Interface/Jobs/ViperHud.cs
+++ b/DelvUI/Interface/Jobs/ViperHud.cs
@@ -130,7 +130,7 @@ namespace DelvUI.Interface.Jobs
             
             var start = new Tuple<PluginConfigColor, float, LabelConfig?>(Config.Vipersight.ComboStartColor, 1, null);
             var endFlank = new Tuple<PluginConfigColor, float, LabelConfig?>(Config.Vipersight.ComboEndFlankColor, 1, null);
-            var endRear = new Tuple<PluginConfigColor, float, LabelConfig?>(Config.Vipersight.ComboEndRearColor, 1, null);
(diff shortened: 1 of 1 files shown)


Zeffuro:
So we're fine with this change but the only thing I'd like to change about it. Instead of having "Flank Positional" and "Rear Positional" I'd prefer if we can have something like Steel Fangs ender or Dread Fangs ender or something. 

But I mostly don't want to start encouraging people to add rear/flank positionals for all the other jobs (especially since DelvCD can do it). But this is a special case because the Vipersight bar is so "unique" as in how it plays for Viper (initially I thought the official bar was useless and just looked at my glowing buttons).

So if you can change those labels (you can suggest any other label name as well as long as it isn't rear/flank) we can probably merge it after.

Blitzkind:
Steel Fangs ender and Dread Fangs ender is as good as any. I'll make that update after I get a break from work.

Zeffuro:
Discussed it again with someone and actually you'll need these to check the flank/rear properly: 

```
var isFlankEnder = Utils.StatusListForBattleChara(player).Any(o => o.StatusId is 3646 or 3645);
var isHindEnder = Utils.StatusListForBattleChara(player).Any(o => o.StatusId is 3647 or 3648);
var isGrimEnder = Utils.StatusListForBattleChara(player).Any(o => o.StatusId is 3649 or 3650);
```

So my suggested naming convention of Dread Fang ender and Steel Fangs ender is kind of incorrect, in this case the buffs are named Hind* and Flank* so we probably should use those instead (at least we won't use rear in that case)

opera_2024-07-10_21-41-27

Blitzkind:
Yep! All good here.

Zeffuro (approved):

PR #1218, opened 2024-07-13
feat(viper): adds invalid states to vipersight

cultbaus:
This is an attempt to fix some missing states I've encountered with the Vipersight gauge. Namely, invalid states caused by whiffing the correct combo key will throw your gauge into a pretty grim state usually resulting in a whiffed positional in addition to an incorrect finisher, and "non-standard" buff refreshes resulting from Dreadwinder or Pit of Dread chains will cause the buff refresh recommendation to fail in some cases, but one of those cases is directly out of our opener.

Also, the refresh condition on the two-part chain seems to be less about which debuff is shorter and more about which two-part you pressed last, so this implementation mimics that to primarily keep the gauge in sync with the in-game combo ants in the event that you should drop a buff or click it off.

This also adjusts a conditional on AOE openers to check for Steel Maw and Dread Maw (we check Dread Maw twice currently) to correctly highlight two-part chains for AOE combos. This is admittedly less convenient than using the buff durations, but more true to the in-game representation of combo information.

Overall this should more closely match the behavior of the combo ants, and I've not been able to desync the in-game hotbar from our beloved Vipersight gauge no matter how many times I goof the rotation.

### DIFF
--- a/DelvUI/Interface/Jobs/ViperHud.cs
+++ b/DelvUI/Interface/Jobs/ViperHud.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using Dalamud.Game.ClientState.JobGauge.Types;
 using Dalamud.Game.ClientState.Objects.SubKinds;
 using DelvUI.Config.Attributes;
 using DelvUI.Helpers;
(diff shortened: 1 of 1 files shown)


Zeffuro:
I think I want to try a slightly different approach since this is getting a bit messy. 
I now have the functionality to check for the combo ants and I think I want to introduce an option to be able to use the buff durations or the ingame representation.

cultbaus:
> I think I want to try a slightly different approach since this is getting a bit messy. I now have the functionality to check for the combo ants and I think I want to introduce an option to be able to use the buff durations or the ingame representation.

Works for me!

Zeffuro:
I have opened a PR #1222 
Could you please pull this in on your end and test it, closing this PR so we can continue discussion over there.

PR #1231, opened 2024-07-18
Add glow when ready to reawaken, move timer to text label

Necroskillz:
An idea to use glow when you have Ready to Reawaken buff. The way it worked previously is changing color when you have buff or 50+ gauge, but you can already do that with partial fill color and its not visible when you have 0 gauge but have the ready buff.

The glow could be used on other jobs that have the "free usage" of something instead of using gauge (like MCH, RPR).

Regarding the Reawakened Timer, i think its pretty useless information to have and the Chunked Bar is not prepared to handle it like that (if partial fill color is on, the timer is grey). Imo makes sense to have a number label like reaper if anyone wants it for some reason, tho i expect most people would just disable it. This also fixed the bug that the toggle to enable it wasnt working.

### DIFF
--- a/DelvUI/Interface/Jobs/ViperHud.cs
+++ b/DelvUI/Interface/Jobs/ViperHud.cs
@@ -11,6 +11,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
+using DelvUI.Enums;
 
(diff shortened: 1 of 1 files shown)


Necroskillz:
actually not sure if this is super needed, can just be solved with DelvCD. gonna close this for now.

PR #1234, opened 2024-07-20
fix(job bars): corrects job bar options

cultbaus:
What: Moves the `Invert` checkbox from `ChunkedBarConfig` -> `VipersightBarConfig`.
Why: The invert checkbox was meant for specifically the Vipersight gauge, as `GrowDirection` satisfies the `Invert` condition for most bars.

What: Confirm `GlowConfig.Enabled` is truthy before glowing `ChunkedBarConfig` elements
Why: The `Glow.Enabled` checkbox in `Vipersight` doesn't do anything in the current implementation.

### DIFF
--- a/DelvUI/Interface/Bars/BarConfig.cs
+++ b/DelvUI/Interface/Bars/BarConfig.cs
@@ -48,10 +48,6 @@ namespace DelvUI.Interface.Bars
         [Order(41)]
         public bool HideWhenInactive = false;
 
-        [Checkbox("Invert", spacing = true)]
-        [Order(44)]
-        public bool Invert = false;
-
         public BarConfig(Vector2 position, Vector2 size, PluginConfigColor fillColor, BarDirection fillDirection = BarDirection.Right)
         {
             Position = position;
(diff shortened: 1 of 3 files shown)


Tischel:
You shouldn't remove a property from the base class `BarConfig`.
Instead add a new specific property with a different name in your custom bar and disable showing the original one by using `[DisableParentSettings("Invert")].

There's many examples of this in the code already for you to check.

cultbaus:
> You shouldn't remove a property from the base class `BarConfig`. Instead add a new specific property with a different name in your custom bar and disable showing the original one by using `[DisableParentSettings("Invert")]`.
> 
> There's many examples of this in the code already for you to check.

I had erroneously introduced this property to the `BarConfig` with #1199 when it should have been on the `VipersightBarConfig` originally, but if you'd prefer to preserve this property on `BarConfig` I am happy to adjust per your feedback.

Tischel:
> > You shouldn't remove a property from the base class `BarConfig`. Instead add a new specific property with a different name in your custom bar and disable showing the original one by using `[DisableParentSettings("Invert")]`.
> > There's many examples of this in the code already for you to check.
> 
> I had erroneously introduced this property to the `BarConfig` with #1199 when it should have been on the `VipersightBarConfig` originally, but if you'd prefer to preserve this property on `BarConfig` I am happy to adjust per your feedback.

Oh sorry, in that case disregard what I said.

PR #1256, opened 2024-08-19
add checkbox to StatusEffectsListConfig to allow showing effects when entity is dead

k0etsu:
currently all status effects are hidden when an entity is dead

however, some debuffs persist through death and being able to see them at a glance on party frames is helpful

00047_ffxiv_dx11_GmfUAPG0K7


Tischel:
Implemented in v2.3.1.7

PR #1260, opened 2024-08-26
Monk: Brotherhood Chakra

vond82:
#### As per request on Discord by Moon and Cathella

- Support for 10 Chakra during Brotherhood.
- Options added to set Brotherhood Fill Color, to Show Glow when full Chakra and to show a Duration label during Brotherhood.

#### How it works:

- When Brotherhood is active, everything will look normal for Chakra 1-5. For Chakras 6-10 the bar will "reset" and start building with the Brotherhood Fill Color instead of the standard Fill Color.
- Once you hit 10 Chakra with the Brotherhood buff or 5 Chakra without the bar will glow if the Show Glow option is enabled.

Below is a simple example with DelvUIs default monk gauge setup. First you see 5 chakra without Brotherhood, so the bar has the glow added. Once Brotherhood is used the glow disappears. Once we use an ability to gain Chakra, it starts building Chakras 6-10 in the Brotherhood Fill Color:

5f0d57023ab541ef7351a9ef926b2203

### DIFF
--- a/DelvUI/Interface/Jobs/MonkHud.cs
+++ b/DelvUI/Interface/Jobs/MonkHud.cs
@@ -102,20 +102,28 @@ namespace DelvUI.Interface.Jobs
                 return;
             }
 
-            bool brotherhoodActive = false;
+            LabelConfig[]? labels = null;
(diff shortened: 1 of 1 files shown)


Zeffuro:
This PR has become very stale, I don't remember if we wanted this or not but if we still want this feel free to reopen or make a new PR with the changes.

PR #1262, opened 2024-09-09
Some requested party cooldowns additions

vond82:
Some requested party cds:

- SCH: Summon Seraph
- DRG:  Geirskogul and Dragonfire Dive
- MNK: Riddle of Wind
- MCH: Barrel Stabilizer
- RDM: Manafication


Zeffuro:
Since there was no reply and it's over 1.5 years old I'm going to close this for now. If we still want these we can always redo them or reopen this.

PR #1285, opened 2025-04-22
fix drk dark missionary to lvl 66 in party cooldowns

k0etsu:
assume it got missed on dt patch 7.1


Zeffuro:
Looking at JobGuide it should be 66 yeah, this should be good to merge
zen_2025-08-20_19-48-34

PR #1293, opened 2025-08-05
Initial API13 Update

Zeffuro:
Allows DelvUI to build.
This is untested ingame yet but fixes all the ImGui bindings and the AtkUnitBase changes that now require the .Address to be passed.

Reminder also that there's still a PR open for fixing the IsHostile check by using the nameplate signature (which may need to be updated but that sig has been stable for like mooooonths now).

If more changes are required I'll update the PR but this should be the brunt of the work.

----

Subtle reminder to also pull in #1291 after merging this.

### DIFF
--- a/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
+++ b/DelvUI/Config/Attributes/ConfigTypeAttributes.cs
@@ -3,7 +3,7 @@ using Dalamud.Interface.Utility;
 using DelvUI.Enums;
 using DelvUI.Helpers;
 using DelvUI.Interface.GeneralElements;
-using ImGuiNET;
+using Dalamud.Bindings.ImGui;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -323,7 +323,7 @@ namespace DelvUI.Config.Attributes
     [AttributeUsage(AttributeTargets.Field)]
     public class InputTextAttribute : ConfigAttribute
     {
-        public uint maxLength;
+        public int maxLength;
         public bool formattable = true;
 
         private string _searchText = "";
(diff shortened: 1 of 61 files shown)


k0etsu:
Just a heads up, not sure the enemy list is functioning correctly with this build.

Zeffuro:
> Just a heads up, not sure the enemy list is functioning correctly with this build.

EnemyList should work now, there was also an instance where PartyMemberInfo was pulling from the wrong array so that should technically be fixed too.

k0etsu:
> PartyMemberInfo

Is that partially why cross world parties would never update with job info like vanilla? Or why party frames always showed alliance a even if you were in another? (For pf alliances specifically)

Zeffuro:
> > PartyMemberInfo
> 
> Is that partially why cross world parties would never update with job info like vanilla? Or why party frames always showed alliance a even if you were in another? (For pf alliances specifically)

The Job Info not updating is another problem by itself that needs to be looked into (not easy to test).
The bug with Alliance A could be solved by that part

k0etsu:
If there is a location to start poking around, where would you suggest? I have wanted to figure these two problems out but haven't really known where to start, obviously not necessary for api13 release but still an issue to pursue.

zbee:
This leaks DelvUI styles out to other plugins whenever you manually switch Profiles, at least.

Zeffuro:
> This leaks DelvUI styles out to other plugins whenever you manually switch Profiles, at least.

I'm testing this but it's not happening for me, I tried both switching from /delvui -> Profiles -> Active Profile dropdown and swapping using a command (/delvui profile [profilename] )

Can you show it happening/check logs if there's anything relevant?

zbee:
Discussing it with Zeffuro in Discord, but will post what I saw here for posterity's sake I suppose:

https://github.com/user-attachments/assets/4909b054-642b-4cd6-bc4b-525a8944f2aa

PR #1295, opened 2025-08-12
Implement player chatbubbles/rework NPC chatbubbles cliprects

Zeffuro:
Implements clipping for player chat bubbles and separate the settings for player chat bubbles and NPC chat bubbles.

https://github.com/user-attachments/assets/ed42ec8f-f784-4b96-b4b7-e4df7ab4a589


DelvUI/Helpers/InputsHelper.cs
```diff
@@ -1,7 +1,7 @@
 ﻿/*
```
Tischel, line 1: All the changes in this file don't seem relevant.

Issue #1290, opened 2025-06-28
[Bug]: Party list does not update correctly from cross-world


k0etsu:
### What are you trying to do?

Use a cross-world party and have it displayed correctly in party frames.

### What is the expected behavior?

Cross-world party should update when party members change jobs, currently not always the case. Also should update to a regular party when instancing into a duty, doesn't always behave like this.

### What actually happened?

Cross-world party does not update member's jobs as it should. Also when instancing into a duty occasionally the party frames still display as if it is a cross-world. Need to manually change party member order to fix.

### Suggested solution

Seems to be some kind of callback not firing correctly when receiving updated info from the client state

### Logs

```shell

```

### FFXIV Update

- [x] I have confirmed that I have the latest version of XIVLauncher and DelvUI.

Zeffuro:
I think this is fully fixed now right?

k0etsu:
Yeah this is fixed

PR #1298, opened 2025-08-17
[Bugfix] Fix cross world/alliance party frames

### LINKED ISSUE #1278: [Bug]: Party Frames show Alliance A regardless of which alliance player is a part of
### What are you trying to do?

Party frames only showing alliance A when I am in C

### What is the expected behavior?

Show me alliance C

### What actually happened?

Only showing alliance A

### Suggested solution

_No response_

### Logs

_No response_

### FFXIV Update

- [X] I have confirmed that I have the latest version of XIVLauncher and DelvUI.

k0etsu:
Should fix #1278 and #1290

### DIFF
--- a/DelvUI/Interface/Party/PartyFramesMember.cs
+++ b/DelvUI/Interface/Party/PartyFramesMember.cs
@@ -3,6 +3,7 @@ using Dalamud.Game.ClientState.Party;
 using DelvUI.Helpers;
 using Dalamud.Bindings.ImGui;
 using System;
+using FFXIVClientStructs.FFXIV.Client.UI.Info;
 
 namespace DelvUI.Interface.Party
 {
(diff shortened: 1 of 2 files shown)


Zeffuro:
Tested for a bit and I can confirm #1290 is fixed
Can't really test #1278 because I don't really do 24man content other than Alliance Roulette but it seems good.

k0etsu:
> Tested for a bit and I can confirm #1290 is fixed Can't really test #1278 because I don't really do 24man content other than Alliance Roulette but it seems good.

I tested by setting up a pf for any 24+ man content and starting in alliance b and having two friends shuffle around in different permutations, doesn't require much more than that I think

Zeffuro:
I pretty much only play with one normally in FFXIV so there's that XD
But the code looks good so it prob works fine

PR #1300, opened 2025-08-20
Rework Visibility to not use ChatHelper

Zeffuro:
Same as the other PR, only pull in when current stg is mainline Dalamud.
Gets rid of ChatHelper fully


Zeffuro:
Oh right this one should be good to go, ApplyHudLayout got added in the last Dalamud apparently

PR #1320, opened 2026-02-17
Fix for party cooldowns not resetting on wipe

### LINKED ISSUE #1319: [Bug]: Party Cooldowns don't reset on wipe
### What are you trying to do?

Noticed recently that party cooldowns are not resetting on wipe

### What is the expected behavior?

Cooldown timers should reset on wipe

### What actually happened?

They are not resetting

[video](https://files.catbox.moe/a1s8rn.mp4)

### Suggested solution

https://github.com/DelvUI/DelvUI/blob/816088706b9b1242959393694a964839ae852888/DelvUI/Interface/PartyCooldowns/PartyCooldownsManager.cs#L153

Something related to this one would assume

### Logs

```shell

```

### FFXIV Update

- [x] I have confirmed that I have the latest version of XIVLauncher and DelvUI.

k0etsu:
Really poorly implemented

Should resolve #1319

### DIFF
--- a/DelvUI/Interface/PartyCooldowns/PartyCooldownsManager.cs
+++ b/DelvUI/Interface/PartyCooldowns/PartyCooldownsManager.cs
@@ -1,6 +1,7 @@
 using Dalamud.Game.ClientState.Conditions;
 using Dalamud.Game.ClientState.Objects.Types;
 using Dalamud.Hooking;
+using Dalamud.Plugin.Services;
 using DelvUI.Config;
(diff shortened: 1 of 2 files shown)


Zeffuro:
Whoops I didn't realize making changes to this branch would automerge it, my bad.

Wasn't as poorly implemented. I just cleaned it up so it uses the same style as the other dalamud services and swapped out the old manual signature for the new one in the process.

